PR #6, opened 2017-07-17
Remove "Open with Firefox"

### LINKED ISSUE #5: Remove "Open with Firefox"
This is needed because Focus does not position itself as a works-for-everything browser. We probably don't need this. Remove it just to make structure cleaner! We can easily roll back if we actually need this (or at least see what needs to be add back from the commit message if the structure is already too dramastically changed).

mTwTm:
Fix #5


mTwTm:
Note that the l10n is not yet settled, so string resources are not yet removed.

PR #65, opened 2017-07-27
Open preference in sheet

walkingice:


walkingice:
@mTwTm , I did a little modification on menu, wish it won't break anything you are working on.

mTwTm:
@walkingice I will be using Linearlayouts instead of Buttons. I think I can cover it and will let you know once I submitted the change. :D

PR #76, opened 2017-07-27
Add a download/history panel based on BottomSheet

mTwTm:
We currently choose to implement with BottomSheetDialog out of simplicity. This may lead to performance issues for not keeping the inflated view. We should investigate if that becomes an issue in the future.


mTwTm:
Fixes #76

PR #123, opened 2017-08-03
Download

WeiWW:
1. Download list layout
2.When user in the download list, download list layout will update when download is completely


Conjuror:
Verified w/ build 143

Issue #109, opened 2017-08-02
Unable to pinch to zoom-in at google map


Conjuror:
as title

mTwTm:
Should work in Build 357

Conjuror:
verified in build 357

PR #126, opened 2017-08-03
Fix Bug #109 - Bypass non-first touch event to super to allow webview …

idiotmax:
…handle pinch to zoom #109


Conjuror:
verified in build 234.

PR #136, opened 2017-08-04
Activity would be landscape if device is in lanscape mode before launch

idiotmax:


Conjuror:
verified in build 214

PR #138, opened 2017-08-05
Update TopSite with async favicon loading for demo

idiotmax:


Conjuror:
nothing to verify

Conjuror:
icons are now displayed in build 214, but it's not the final version.

PR #201, opened 2017-08-11
Notify Media Scanner after screenshot is captured.

### LINKED ISSUE #198: Long page screenshot is not immediately shown in the Galleries.
Screenshots do not show in Galleries immediately.
We're not notifying the system yet. We need to actively let system know of the existence to speed up.

mTwTm:
Fixes #198


mTwTm:
The implementation is incorrect, patch in #225.

Issue #131, opened 2017-08-04
back key should work even we have home screen in track


Conjuror:
While browsing, back key should work to visit previous web page. If we attempt to visit home screen during browsing, it should be recorded as well.

i.e,
A→B→C→(software home key) Top Sites→D→E
press back key will visit following pages in order:
D→Top Sites→C→B→A

mTwTm:
This can be even more complex when "pages open through other apps" get into play. Does the history reset or not under such situation should be discussed too.

Conjuror:
Verified on build 382.

PR #213, opened 2017-08-12
Fragment management

walkingice:
Rewrite fragment management logic to match design spec


walkingice:
This PR tries to implement fragment stack logic which defined by our spec.

walkingice:
Implement logic of fragment stack management. Should also fix #131

PR #434, opened 2017-08-28
fix #390 : Enable DOM storage and can delete all storage data when cl…

### LINKED ISSUE #390: Unable to load https://www.mop.com.tw/linkou/tw/index.html
Focus loads fine

WeiWW:
Enable DOM storage and can delete all storage data when cl…


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -340,6 +341,8 @@ private void onScreenshotsClicked() {
     }
 
     private void onDeleteClicked() {
+        WebStorage.getInstance().deleteAllData();
```
mTwTm: This should also contribute to the diff in line 347

Issue #247, opened 2017-08-14
Contextual hint to introduce feature when user doing android screenshot


chianger:
UX Spec v1.8, page37 has the details

chianger:
Labeled UX/VD to request visual design

harly21:
Carol released the visual spec here: https://drive.google.com/drive/folders/0B-9PIePlQZRlYlZTdDlLNXRwUFU

dakaizou:
Screenshot onboarding doesn't show up until granting storage permission

Build: 1486
DuT: Samsung

dakaizou:
verified in build 1512

PR #479, opened 2017-08-30
Show screenshot on boarding dialog.

joseph-yar:
- feature 10504


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -809,6 +838,42 @@ private void notifyNewScreenshot(String path) {
...
+    private void showScreenshotOnBoardingDialog() {
+        if(dialogScreenshotOnBoarding != null && dialogScreenshotOnBoarding.isShowing())
+            return;
+
+        if(getActivity() != null && Settings.getInstance(getActivity()).shouldShowScreenshotOnBoarding()) {
+            dialogScreenshotOnBoarding = new AlertDialog.Builder(getActivity()).create();
+            View dialogView = LayoutInflater.from(getActivity()).inflate(R.layout.layout_screenshot_onboarding_dialog, null);
+
+
+            TextView textView = (TextView) dialogView.findViewById(R.id.dialog_screenshot_on_boarding_text);
```
idiotmax: Pull view implementation detail to other class

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -132,14 +138,37 @@ public static BrowserFragment create(String url) {
...
 
+    // monitor screenshot is created
+    private FileObserver fileObserver;
+
+    private AlertDialog dialogScreenshotOnBoarding;
+
     @Override
     public void onPause() {
         super.onPause();
+        if(fileObserver != null)
```
idiotmax: always quote with {} in code block

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -86,6 +91,7 @@
     private static final int ANIMATION_DURATION = 300;
     private static final String ARGUMENT_URL = "url";
     private static final String RESTORE_KEY_DOWNLOAD = "download";
+    private static final String SCREENSHOT_PATH = Environment.getExternalStorageDirectory() + File.separator + Environment.DIRECTORY_PICTURES + File.separator + "Screenshots" + File.separator;
```
mTwTm: Samsung is using a different directory for screenshots, we should either watch a set of directories or try to guess where the screenshot directory is with the device info.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -132,14 +138,37 @@ public static BrowserFragment create(String url) {
...
     public void onPause() {
         super.onPause();
+        if(fileObserver != null)
+            fileObserver.stopWatching();
     }
 
     @Override
     public void onResume() {
         super.onResume();
+        if(Settings.getInstance(getActivity()).shouldShowScreenshotOnBoarding()) {
```
mTwTm: We need to ask for External Path permission to do this

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mTwTm: Is there any concern to not use
`Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES) + File.separator + "Screenshots" + File.separator` here?

mTwTm: Let's start with listening to the root and on every event check if the path has a "screenshot" in it (case insensitive, does not check for /screenshot/ since maybe a folder could have the name screenshots screenshot_ ... etc to cover most of the cases. Let's ignore if there are some devices that have a different name for screenshot folders for now. We'll file another bug if such device exist.) Listening to the root may raise some performance issue but let's start with the easiest one and see if it works. 😃

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mTwTm: After discussion, we decide to not ask for permission for now so nothing to change here.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mTwTm: This is just a probably slightly more promising way to get the correct but same directory. It does not matters anymore now anyway if we're not listening to this. 😜

app/src/main/java/org/mozilla/focus/components/RecursiveFileObserver.java
```diff
@@ -0,0 +1,77 @@
+package org.mozilla.focus.components;
+
+import android.os.FileObserver;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
+public class RecursiveFileObserver extends FileObserver {
```
mTwTm: I should have recalled that FileObservers are not recursive before saying that my proposal is the easy way... sorry.
This does not work on my LG and I believe doing this recursively on every folder is not a good idea. I'm thinking of listening to MediaStore.Images.Media.EXTERNAL_CONTENT_URI with a ContentObserver. What do you think, @idiotmax ?

mTwTm: We're actually implementing screenshot observing with the same concept on Fennec, amazing! https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/ScreenshotObserver.java Will dig deeper tomorrow.

mTwTm: I've checked on my LG with the old single folder implementation, it does not work. It's not sending the CREATE event, but only OPEN, ACCESS, CLOSE_NOWRITE. My opinion is still toward changing the architecture since I've heard some unreliability on FileObserver after Marshmallow.

joseph-yar:
Need to confirm when to check READ_EXTERNAL_STORAGE permission.

joseph-yar:
#247

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -964,4 +989,16 @@ private Intent createChooserIntent(WebChromeClient.FileChooserParams params) {
             return FilePickerUtil.getFilePickerIntent(getActivity(), title, mimeTypes);
         }
     }
+
+    @Override
+    public void onScreenshotTaken(String screenshotPath, String title) {
```
mTwTm: My personal feeling is declaring a new type in FragmentListener.TYPE.SHOW_SCREENSHOT_HINT will be cleaner then current implementation.

We won't need a lot of getActivity() in this case, and another benefit would be we can collect all Dialog showing logic there (a simple call to DialogUtils.showScreenshotOnBoardingDialog(this);)

I'm particularly against this because everytime we call getActivity() it can suddenly return null. Keep all logic in the activity seems just more right to me. I'm not 100% certain on this so let me know if you think otherwise.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mTwTm: This is still an issue but I think we can merge first and worry about this if our UX tried and would like to prompt for the permission once we starts the app. Prompt every time we are notified for a new entry does not make sense for me so I guess the only place reasonable will be when the app is started, that we be pretty simple if we want to do so so let's merge this first.

Issue #476, opened 2017-08-30
[Smoke Test] Downloads thumbnails not working


dakaizou:
Downloads thumbnails are total black

Build: 623
DuT: Samsung J7 Prime, advan, andromax

Conjuror:
looks good on build 636

Issue #369, opened 2017-08-22
[dogfooding] default image holder while no download files


Conjuror:
as title

bulid 461

chianger:
can leverage outsourcing resource

Conjuror:
Visual work is done. Remove UX&VD label and wait for engineering work.

cssjh:
**Verified** on build 702

PR #481, opened 2017-08-30
Ann/download

WeiWW:
1.fix issue #369 
2. fix issue #476


app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -17,8 +18,10 @@
 import org.mozilla.focus.utils.IntentUtils;
 
 import java.io.File;
+import java.lang.reflect.Array;
```
mTwTm: We don't need this

mTwTm:
Squashed 1d8da46 and 03671f6 into 6d91dfc and fixed some very minor problems.

Issue #375, opened 2017-08-22
Update send usage data string in setting


joechengla:
Refer to https://bugzilla.mozilla.org/show_bug.cgi?id=1390689
Max and Jack has access to this bug

Conjuror:
@chianger no sure which category you prefer. remove bug to avoid pollute the pool

chianger:
yes. let's track in Zerda V1.0.

Conjuror:
How should I verify access to Telemetry?

Conjuror:
This will be hard for QA to verify. Leave it open as low priority

PR #486, opened 2017-08-30
custom telemetry preference layout - feature #375

Ben-YPLai:
Update send usage data string in setting #375

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/TelemetrySwitchPreference.java
+++ b/app/src/main/java/org/mozilla/focus/widget/TelemetrySwitchPreference.java
@@ -7,13 +7,11 @@ package org.mozilla.focus.widget;
 
 import android.content.Context;
 import android.content.Intent;
-import android.content.res.TypedArray;
-import android.graphics.Color;
-import android.graphics.Paint;
-import android.graphics.drawable.Drawable;
 import android.preference.Preference;
 import android.util.AttributeSet;
+import android.view.LayoutInflater;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.CompoundButton;
 import android.widget.Switch;
 import android.widget.TextView;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/widget/TelemetrySwitchPreference.java
```diff
@@ -41,13 +39,20 @@ public TelemetrySwitchPreference(Context context, AttributeSet attrs, int defSty
     }
 
     private void init() {
-        setWidgetLayoutResource(R.layout.preference_telemetry);
+//        setWidgetLayoutResource(R.layout.preference_telemetry);
```
idiotmax: remove the unused commented code

app/src/main/java/org/mozilla/focus/widget/TelemetrySwitchPreference.java
```diff
@@ -41,13 +39,20 @@ public TelemetrySwitchPreference(Context context, AttributeSet attrs, int defSty
...
 
         // We are keeping track of the preference value ourselves.
         setPersistent(false);
     }
 
     @Override
+    protected View onCreateView(ViewGroup parent) {
+        super.onCreateView(parent);
+        LayoutInflater li = (LayoutInflater)getContext().getSystemService( Context.LAYOUT_INFLATER_SERVICE );
+        return li.inflate( R.layout.preference_telemetry, parent, false);
```
idiotmax: What's the difference if we replace the layout reference in settings.xml from focus_preference to this preference_telemetry? This override could be removed completely, correct?

Ben-YPLai: I think replace the layout reference in settings.xml from focus_preference is just replace a part of base preference layout. I will try which one is correct.

PR #514, opened 2017-08-31
Max/feature/screenshot animation

idiotmax:


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotCaptureTask.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+public class ScreenshotCaptureTask extends AsyncTask<Object, Void, String> {
+
+    private final Context context;
+
+    public ScreenshotCaptureTask(Context context) {
```
mTwTm: I guess Context.getApplicationContext() would make things easier if you want the ApplicationContext

mTwTm: This is invalid, sorry!

Issue #382, opened 2017-08-23
[Smoke Test] download placeholder will appear while tapping download tab quickly


Conjuror:
build 471

as title

If there's a downloaded file in the list, tapping download tab repeatedly will be able to see the placeholder for empty list.

joseph-yar:
#538

chianger:
fixed already

dakaizou:
verified in build 790

Issue #269, opened 2017-08-15
[dogfooding] History panel blinks if a user taps the tab


Conjuror:
already on browsing history panel=>tap on "history", the history list blinks

build 357

dakaizou:
Still happening in build 623

dakaizou:
Verified in build 751

PR #538, opened 2017-09-01
Fix download placeholder will appear while tapping download tab quickly.

joseph-yar:
- issue 471

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/ListPanelDialog.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/ListPanelDialog.java
@@ -28,6 +28,7 @@ public class ListPanelDialog extends DialogFragment {
     private View downloadsTouchArea;
     private View historyTouchArea;
     private View screenshotsTouchArea;
+    private boolean firstLaunch = true;
 
     public static ListPanelDialog newInstance(int type) {
         ListPanelDialog listPanelDialog = new ListPanelDialog();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/fragment/ListPanelDialog.java
```diff
@@ -28,6 +28,7 @@
     private View downloadsTouchArea;
     private View historyTouchArea;
     private View screenshotsTouchArea;
+    private int type;
```
mTwTm: We don't need this, we can use getArguments().getInt(TYPE). getArguments().getInt(TYPE) is also more promising if the Fragment is recycled and recreated by the system.

app/src/main/java/org/mozilla/focus/fragment/ListPanelDialog.java
```diff
@@ -123,18 +124,27 @@ private void setSelectedItem(int selectedItem) {
     }
 
     private void showDownloads() {
-        setSelectedItem(TYPE_DOWNLOADS);
-        showPanelFragment(DownloadsFragment.newInstance());
+        if(type != TYPE_DOWNLOADS) {
```
mTwTm: Let's use getArguments().getInt(TYPE) and don't save a duplicate.

mTwTm:
Fixes #269

app/src/main/java/org/mozilla/focus/fragment/ListPanelDialog.java
joseph-yar: If the entry point for tap download tab is MainActivity.showListPanel() -> ListPanelDialog.newInstance() ->ListPanelDialog.showDownloads(), "if(getArguments().getInt(TYPE) != TYPE_DOWNLOADS)" will not be satisfied. So I need to use "type"(default value is 0) as a local reference to cover this case.

mTwTm: Ah I see. How about solving it by introducing a member variable `boolean firstLaunch = false` to avoid duplicate entries:

```
    private boolean firstLaunch = true;
    private void showDownloads() {
        if ( firstLaunch || getArguments().getInt(TYPE) != TYPE_DOWNLOADS) {
            setSelectedItem(TYPE_DOWNLOADS);
            showPanelFragment(DownloadsFragment.newInstance());
        }
    }
    private void toggleSelectedItem() {
        firstLaunch = false;
    ......
    }
```

Your code will definitely work too so it's only my personal preference to avoid introducing duplicates (Which makes future developers hard to figure out which to update. )

Let me know if you like yours more and I will merge current implementation.

mTwTm: A maybe even better way would be differentiating 1. firstLaunch and 2. Taps by user:
```
    private void showDownloads(boolean firstLaunch) {
        if ( firstLaunch || getArguments().getInt(TYPE) != TYPE_DOWNLOADS) {
            setSelectedItem(TYPE_DOWNLOADS);
            showPanelFragment(DownloadsFragment.newInstance());
        }
    }
```
With this we can even get rid of introducing a new variable but it is slightly confusing the future developers. Personally I like this one the most.

Conjuror:
fixed #382

Issue #349, opened 2017-08-22
Top sites reset along with browsing history clear


Conjuror:
History = Top site => History entry is deleted when top site removed, and when history is cleared, top site is cleared

Conjuror:
P0 to have UX input information to implement

Conjuror:
Count by visit count.

dakaizou:
verified in build 790

PR #547, opened 2017-09-02
Top sites reset along with browsing history clear #349

Ben-YPLai:


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -261,9 +261,22 @@ private BrowserFragment getVisibleBrowserFragment() {
         }
     }
 
+    private Fragment getTopHomeFragment() {
```
idiotmax: Add Nullable annotation

PR #601, opened 2017-09-05
Fix time on History/Screenshot panel should be all caps.

### LINKED ISSUE #545: Time on History/Screenshot panel should be all caps
It's "TODAY" not "Today" according to the VI specs.
**Build:** 653

joseph-yar:
- Closes #545


idiotmax:
Spec updated, keep current design.

Issue #429, opened 2017-08-28
Disable `Search Privately` or at least change the string


mTwTm:
Like focus, we're showing `Search Privately` in the context menu after text selection. We either disable this or change the text to not confuse our users.

![device-2017-08-28-145402](https://user-images.githubusercontent.com/6755074/29762374-caeb4e44-8c00-11e7-82af-3d38618d4465.png)

idiotmax:
@harly21: How about "Turbo Search"

harly21:
After discussing with the team, it is decided to disable "Search Privately"

Conjuror:
[Triage] UX to decide terminology for it

harly21:
Suggest string to be modified to "Search in Firefox Rocket"

idiotmax:
Closed by #703

Conjuror:
build 972

PR #638, opened 2017-09-07
Fix disable `Search Privately` or at least change the string

joseph-yar:
- Closes issue #429


app/src/main/res/values/strings.xml
```diff
@@ -45,7 +45,7 @@
 
     <!-- This is the label of an action we offer when the user selects text in other third-party apps.
          Clicking this action will launch Focus and perform a search in the default search engine. -->
-    <string name="text_selection_search_action">Search Privately</string>
+    <string name="text_selection_search_action">Search in Firefox Rocket</string>
```
idiotmax: Can we use the app_name instead fixed name here?

PR #653, opened 2017-09-07
Check if the position clicked is within the correct range.

### LINKED ISSUE #430: Buddybuild crash report on build #540
Buddybuild detected a crash
`app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java` line `149`

```java


    @Override
    public void onClick(View v) {
        final int position = mRecyclerView.getChildLayoutPosition(v);
        if (position != RecyclerView.NO_POSITION) {
            Object item = mItems.get(position); <-- CRASHED
            if (item instanceof Site) {
                // TODO: modify the following code accordingly
                Intent intent = new Intent(mContext, MainActivity.class);
                intent.setAction(Intent.ACTION_VIEW);
                intent.setData(Uri.parse(((Site) item).getUrl()));
```

[View File](https://github.com/mozilla-tw/Zerda/blob/b59070c69cedc38883ca4408c0992954f71a2728/app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java#L149) | [View Full Crash Details](https://dashboard.buddybuild.com/apps/594b679a6ed87e0001fca2cd/crashreport/8aac4e896d41989f69f7779d50a7bb59)

hartsu:
Closes #430

### DIFF
--- a/app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
@@ -154,8 +154,8 @@ public class HistoryItemAdapter extends RecyclerView.Adapter<RecyclerView.ViewHo
 
     @Override
     public void onClick(View v) {
-        final int position = mRecyclerView.getChildLayoutPosition(v);
-        if (position != RecyclerView.NO_POSITION) {
+        final int position = mRecyclerView.getChildAdapterPosition(v);
+        if (position != RecyclerView.NO_POSITION && position < mItems.size()) {
             Object item = mItems.get(position);
             if (item instanceof Site) {
                 // TODO: modify the following code accordingly


app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
```diff
@@ -155,7 +155,7 @@ public int getItemCount() {
     @Override
     public void onClick(View v) {
         final int position = mRecyclerView.getChildLayoutPosition(v);
```
idiotmax: Would it be better if we use "getChildAdapterPosition"?

hartsu: That would be better. I would revise "getChildLayoutPosition" to "getChildAdapterPosition" and force push it.

PR #658, opened 2017-09-07
fix #448: when delete downloaded file from Zerda, will notify system …

### LINKED ISSUE #448: Delete downloaded file from Zerda, should notify DownloadManager
We store the same download-id from DownloadManager, if we open download-list of Zerda and remove one file from List, we should notify DownloadManager[1]

[1] https://developer.android.com/reference/android/app/DownloadManager.html#remove(long...)

WeiWW:
…DownloadManager


mTwTm:
@walkingice can you help review this? I'm not familiar with our `RelocateService`

mTwTm:
This is already merged

PR #659, opened 2017-09-07
fix #151: can cancel download when it is running.

### LINKED ISSUE #151: [Download] Download should have a cancel button
Currently, only Android 7.0 and above provide cancel button in system tray. We'll need a cancel mechanism for Android 5.x and 6.x.

WeiWW:


app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -85,6 +85,11 @@ private void delete(int position){
         remove(position);
     }
 
+    private void cancel(int position){
+        DownloadManager manager = (DownloadManager) mContext.getSystemService(Context.DOWNLOAD_SERVICE);
+        manager.remove(mDownloadInfo.get(position).getDownloadId());
```
mTwTm: Also remove it from our own DB so that we won't see a "UNKNOWN" item after it is canceled. Or maybe we can add another canceled state which can only be removed but not deleted. I have pinged @FlyTori for further clarification.

mTwTm: Let's remove the item when it is deleted.

WeiWW: So when user cancel the download, we won't record in our DB?

mTwTm: I think whenever we see a STATUS_FAILED let's remove the item directly for simplicity. If you feel like leaving the item there and showing a canceled status would be simpler/better, just let me know and I'll consult @FlyTori . (Or contact here directly :3)

mTwTm: Update: After tracing the code, it actually shouldn't be STATUS_FAILED. If a download is canceled, it will not have an entry in the DownloadManager so we should remove it if we failed to query. See: https://github.com/mozilla-tw/Zerda/blob/2efc0e015a8195077fb8925785f2e5a9822bc176/app/src/main/java/org/mozilla/focus/download/DownloadInfoManager.java#L255

app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -88,6 +88,8 @@ private void delete(int position){
     private void cancel(int position){
         DownloadManager manager = (DownloadManager) mContext.getSystemService(Context.DOWNLOAD_SERVICE);
         manager.remove(mDownloadInfo.get(position).getDownloadId());
+
+        remove(position);
```
mTwTm: This only removes items canceled inside out app, but not items via Notifications.

WeiWW:
@mTwTm , I has squashed

app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
mTwTm: resolved

PR #666, opened 2017-09-07
Remove TurboMode caching from MainActivity, Closes #652

### LINKED ISSUE #652: Turning off Turbo Mode in onboarding doens't work
as title
Build 790

STR:
1. Clear App DATA
2. Launch Zerda
3. Tap Next and turn off Turbo Mode
4. Finish Onboarding pages
5. Tap on Menu

Expected Result:

3. Turbo mode is off
5. Turbo mode is off

Actual Result:

5. Turbo mode is on

idiotmax:


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -268,12 +266,18 @@ private void updateMenu() {
         final BrowserFragment browserFragment = getVisibleBrowserFragment();
         final boolean hasLoadedPage = browserFragment != null && !browserFragment.isLoading();
         final boolean canGoForward = browserFragment != null && browserFragment.canGoForward();
+
+        menu.findViewById(R.id.menu_turbomode).setSelected(isTurboEnabled());
```
mTwTm: We don't need to do this here, let's move this to onSharedPreferenceChanged() since updateMenu is called everytime a page load is finished or menu is opened, that's an overkill for this situation.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -246,7 +244,7 @@ private void setUpMenu() {
         captureButton = menu.findViewById(R.id.capture_page);
         refreshIcon = menu.findViewById(R.id.action_refresh);
         stopIcon = menu.findViewById(R.id.action_stop);
-        menu.findViewById(R.id.menu_turbomode).setSelected(mTurboModePref);
+        menu.findViewById(R.id.menu_turbomode).setSelected(isTurboEnabled());
```
mTwTm: Suggest to do the same (remove the caching) for blockimage too, just to keep just one logic for similar things which hopefully makes the life of future developers easier...

PR #688, opened 2017-09-08
Fix after deleting screenshot image, screenshot preview still exists …

### LINKED ISSUE #642: After deleting screenshot image, screenshot preview still exists on "My shots"
Actual result: 
1. delete screenshot image
2. return to "My shots" panel
3. the deleted screenshot preview still on the panel

Expected result
1. delete screenshot image
2. return to "My shots" panel
3. the deleted screenshot should be removed from "My shots" panel

Build: 0.4.790.beta

joseph-yar:
…on "My shots"

- Closes #642

### DIFF
--- a/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotItemAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotItemAdapter.java
@@ -26,7 +26,7 @@ import java.util.List;
  */
 
 public class ScreenshotItemAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> implements View.OnClickListener,
-        QueryHandler.AsyncQueryListener, QueryHandler.AsyncDeleteListener {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotItemAdapter.java
```diff
@@ -161,19 +161,16 @@ public void onQueryComplete(List result) {
...
 
-    @Override
-    public void onDeleteComplete(int result, long id) {
-        if (result > 0) {
-            if (id < 0) {
-                final int count = mItems.size();
-                mItems.clear();
-                notifyItemRangeRemoved(0, count);
+    public void onItemDelete(long id) {
+        if (id < 0) {
```
mTwTm: The id should always be >=0 since we're not feeding -1 like we do in clear_browsing_history. Should be safe to just do the things in else {} here.

PR #694, opened 2017-09-08
Add HomePanel swipe gesture

idiotmax:
Swipe down to focus on Urlbar
Swipe up to show menu panel

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/SwipeMotionDetector.java
+++ b/app/src/main/java/org/mozilla/focus/utils/SwipeMotionDetector.java
@@ -1,0 +1,66 @@
+package org.mozilla.focus.utils;
+
+import android.content.Context;
+import android.view.GestureDetector;
+import android.view.MotionEvent;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/utils/SwipeMotionDetector.java
```diff
@@ -0,0 +1,66 @@
...
+        public boolean onDown(MotionEvent e) {
+            return true;
+        }
+
+        @Override
+        public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY) {
+            boolean result = false;
+            float diffY = e2.getY() - e1.getY();
+            float diffX = e2.getX() - e1.getX();
+            if (Math.abs(diffX) > Math.abs(diffY)) {
```
mTwTm: We will honor left/right swipe over up/down swipe if we swipe diagonal, not sure if this is expected and should be documented.

Issue #58, opened 2017-07-26
[Function] Setting->Clear browsing data


chianger:

Conjuror:
Not sure why to close this issue. It's not implemented yet on build 471 (Daily build on Aug 23rd), neither see any commits related to it by now.

Would you mind to point out the related commits if I missed it, @idiotmax ? Thank you for kindness.

chianger:
Able to clear "browsing history", "cookies", "cache".
A snackbar will show up after user clears data.

chianger:
needs UX (Jack/Tori) to confirm the snack bar message.

Jacklin-ux:
The snackbar message would be "Browsing data was cleared", when Cache option is unchecked.
I will update this to the spec

chianger:
The snackbar message would be "Browsing data was cleared", (P1)
; when Cache option is unchecked, I will update this to the spec (P2)

chianger:
Looks like it's working already!

mTwTm:
We still need remove cache implementation

chianger:
as removing cache is merged.

Conjuror:
Build 972

PR #702, opened 2017-09-11
add setting clear cache function

Ben-YPLai:
Setting -> clear cache  #58

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/Settings.java
+++ b/app/src/main/java/org/mozilla/focus/utils/Settings.java
@@ -10,6 +10,7 @@ import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/utils/Settings.java
```diff
@@ -115,4 +116,17 @@ public void setDefaultSearchEngine(SearchEngine searchEngine) {
     /* package */ String getPreferenceKey(int resourceId) {
         return resources.getString(resourceId);
     }
+
+    /**
+     * clear cache and return clear msg
+     * @param context
+     * @return clearCache Msg
+     */
+    public String clearCache(Context context) {
```
mTwTm: Let's simply return the long diff and let view decide how it wants to format/show it.

mTwTm: I'm also thinking if we should change to static method if we don't need the class instance.

PR #706, opened 2017-09-11
add setting clear cache function

Ben-YPLai:
Setting->Clear browsing data -> Clear Cache #58


app/src/main/java/org/mozilla/focus/utils/Settings.java
```diff
@@ -115,4 +116,9 @@ public void setDefaultSearchEngine(SearchEngine searchEngine) {
     /* package */ String getPreferenceKey(int resourceId) {
         return resources.getString(resourceId);
     }
+
+    public static long clearCache(Context context) {
```
mTwTm: Let's put this into FileUtils

PR #725, opened 2017-09-12
Change DownloadInfo DB version. Since remove file name column and add…

WeiWW:
… file path column, so need to update db version.


mTwTm:
We will not upgrade the db

PR #737, opened 2017-09-12
add screenshots for l10n

Conjuror:
as title. add screenshots for l10n reviewing


Conjuror:
After discussion, keep the screenshots local

PR #750, opened 2017-09-13
Fix deleting 4 top sites make other elements slide down a little on s…

### LINKED ISSUE #670: [Smoke Test] Deleting 4 top sites make other elements slide down a little on small devices
Related to #640, not very salient but users can tell the movement

screenshot
screenshot

Build: 855
DuT: Samsung, Advan, Andromax

joseph-yar:
…mall devices

- Closes #670

### DIFF
--- a/app/src/main/res/layout/fragment_homescreen_item_recyclerview.xml
+++ b/app/src/main/res/layout/fragment_homescreen_item_recyclerview.xml
@@ -4,7 +4,7 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@id/main_list"
     android:layout_width="296dp"
-    android:minHeight="186dp"
+    android:minHeight="186.67dp"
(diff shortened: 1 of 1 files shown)


joseph-yar:
I reproduced this issue on hdpi(1.5x) device. When there are 2 rows in recyclerview(@id/main_list), the height of recyclerview is 280px. If there's only one row(less than five top sites), the height becomes 279px. I believe this is the root cause. Therefore the minimum height should be 280px (i.e. 280/1.5=186.67 dp), the original minHeight is 186dp(i.e. 186*1.5=279px).

mTwTm:
Thanks for the update, that should be the reason I failed to reproduce on my mdpi/xhdpi devices.I was calculating the sum of the label and expecting things to be the same, so seems like there are some extra px due to rounding (1.5x). Thanks!

Issue #120, opened 2017-08-02
Unable to use OAuth login


Conjuror:
> Dogfooding comment:
> "cannot login to instragram using Facebook login (already logged into facebook)
> Quora is the same, but we can login with Google (interesting!)"

as title

chianger:
added to sprint6 to investigate first.

walkingice:
compare to other browser, if we add this line, Instagram oauth login seems start working

```java
CookieManager.getInstance().setAcceptThirdPartyCookies(webkitView, true);
```

Unfortunately, the method is involved since API 23 :-/

mTwTm:
If we're using this, we should probably mention this in the security review too.

Ben-YPLai:
Apps that target KITKAT or below default to allowing third party cookies. Apps targeting LOLLIPOP or later default to disallowing third party cookies.

cssjh:
**Verified** on 1050.

PR #761, opened 2017-09-13
logic: WebkitView accept third party cookies

Ben-YPLai:
issue  Unable to use OAuth login #120

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
+++ b/app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
@@ -13,6 +13,7 @@ import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -65,6 +65,7 @@ public WebkitView(Context context, AttributeSet attrs) {
 
         linkHandler = new LinkHandler(this);
         setOnLongClickListener(linkHandler);
+        CookieManager.getInstance().setAcceptThirdPartyCookies(this, true);
```
walkingice: If this API is involved since API 21, should we add Api level checking?
```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
   ....
}
```

And `WebViewProvider.configureDefaultSettings` would be better place to do our webview-configuration.

PR #778, opened 2017-09-14
Enable image loading for about/error page and disable it after page loaded if needed.

### LINKED ISSUE #736: [dogfooding] Block image will block error page and about page as well
build 950.

hartsu:
Closes #736

### DIFF
--- a/app/src/main/res/raw/about.html
+++ b/app/src/main/res/raw/about.html
@@ -11,5 +11,6 @@
 <img src="%wordmark%" id="wordmark"/>
 <p class="subtitle">%about-version%</p>
 %about-content%
+<input type="hidden" id="mozillaAboutPage">
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


hartsu:
Using `webView.getSettings().setLoadsImagesAutomatically(true);` in `InfoActivity.onCreateView()` would solve this issue only for launching about page via setting.

Launching about page via entering "focusabout:" directly and launching error page will still have problem.

mTwTm:
Yeah you're totally right. I missed that point. I'm hesitant here because we're relying on the paring of begin loading and onPageFinished while this may not always be the case -- User may start error page / about page but leave before onPageFinished is fired, then we failed to turn auto loading off. We can do this for every onPgeFinished (abandoning the isAboutPage or isErrorPage check) to avoid this but this may lead to performance regression, and image will still be loaded for -- at least once. However I can't think of a better solution either without changing the architecture intensively, @idiotmax can you give some comment here?

PR #789, opened 2017-09-15
logic: fix hard coded string

Ben-YPLai:


app/src/main/res/values/strings.xml
```diff
@@ -274,4 +274,7 @@
     <string name="message_cleared_browsing_data">Browsing data cleared</string>
     <string name="message_deleted_screenshot">Screenshot deleted</string>
     <string name="message_cannot_find_screenshot">Cannot display the screenshot</string>
+
+    <!-- Remove Top site item -->
+    <string name="remove_topsite">Remove</string>
```
idiotmax: How about making the string ID more generic? Try "remove"?

PR #795, opened 2017-09-15
logic: Improve title guessing from domain

### LINKED ISSUE #624: Improve title guessing from domain
We don't want to show just "h" when we only for domain. We should try to guess the most important part.

Ben-YPLai:
close #624


app/src/main/java/org/mozilla/focus/home/TopSiteAdapter.java
```diff
@@ -85,7 +85,7 @@ public int getItemCount() {
     }
 
     public void addSite(int index, @NonNull Site toAdd) {
-       this.sites.add(index, toAdd);
+        this.sites.add(index, toAdd);
```
mTwTm: Personally I would move this to another commit so this would not pollute this commit, but it's just a matter of preference.

PR #870, opened 2017-09-20
Remove redundant code when get fav icon.

joseph-yar:


app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -272,19 +272,11 @@ public void onReceiveValue(String errorPage) {
                             return;
                         }
 
-                        String title = getTitle();
-                        final Bitmap favIcon;
-                        if (TextUtils.isEmpty(title)) {
-                            favIcon = null;
```
mTwTm: Shall we remove this too?
https://github.com/mozilla-tw/Zerda/blob/15ab1bd95302d15eafc54ebec2c66157aee451a9/app/src/main/java/org/mozilla/focus/home/TopSiteAdapter.java#L58

joseph-yar: Yes, we shall.

Issue #614, opened 2017-09-06
[Smoke Test] Shows "No History/Downloads" when loading items


cssjh:
Should be showing "loading"? It's a little bit confusing since there still are some items.
**Build:** 751

mTwTm:
Fixed in #924

cssjh:
**Verified** on 1412.

PR #924, opened 2017-09-25
When opening the download panel, don't show "No downloads"

WeiWW:
1. as title
2.Can load more when scroll the recyclerview

### DIFF
--- a/app/src/main/java/org/mozilla/focus/history/BrowsingHistoryFragment.java
+++ b/app/src/main/java/org/mozilla/focus/history/BrowsingHistoryFragment.java
@@ -26,6 +26,9 @@ import org.mozilla.focus.telemetry.TelemetryWrapper;
 
 public class BrowsingHistoryFragment extends PanelFragment implements View.OnClickListener, HistoryItemAdapter.HistoryListener {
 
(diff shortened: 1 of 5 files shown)


mTwTm:
First commit seems good to me.

app/src/main/java/org/mozilla/focus/fragment/DownloadsFragment.java
```diff
@@ -39,6 +39,22 @@ public static DownloadsFragment newInstance() {
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         recyclerView = (RecyclerView) inflater.inflate(R.layout.fragment_downloads, container, false);
+
+        recyclerView.addOnScrollListener(new RecyclerView.OnScrollListener() {
+            @Override
+            public void onScrolled(RecyclerView recyclerView, int dx, int dy) {
+                super.onScrolled(recyclerView, dx, dy);
```
mTwTm: The onScrolled is fired infinitely, we shall solve this (which should be on my side) first before we move on.

mTwTm:
Fixes #614

PR #932, opened 2017-09-25
Alternative solution to #736

### LINKED ISSUE #736: [dogfooding] Block image will block error page and about page as well

idiotmax:
Closes #736


app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
```diff
@@ -112,9 +113,11 @@ public void onPageFinished(WebView view, final String url) {
 
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, String url) {
+        view.getSettings().setLoadsImagesAutomatically(true);
         if (url == null) {
             // in case of null url, we won't crash app in release build
             if (AppConstants.isReleaseBuild()) {
+                view.getSettings().setLoadsImagesAutomatically(!Settings.getInstance(view.getContext()).shouldBlockImages());
```
hartsu: This seems not necessary because the LoadsImagesAutomatically setting has no effect while loading url "", and it will be reseted while loading the next url.

Issue #974, opened 2017-09-27
Need loadmore for download panel


mTwTm:

dakaizou:
verified in build 1374

PR #981, opened 2017-09-27
Improve download panel loadmore

WeiWW:
as title

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/DownloadsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/DownloadsFragment.java
@@ -76,7 +76,9 @@ public class DownloadsFragment extends PanelFragment implements DownloadInfoMana
 
     @Override
     public void tryLoadMore() {
-        // TODO: 2017/9/26
+        if (!mDownloadListAdapter.isLoading() && !mDownloadListAdapter.isLastPage()){
+            mDownloadListAdapter.loadMore();
+        }
     }
 
     @Override
@@ -104,7 +106,7 @@ public class DownloadsFragment extends PanelFragment implements DownloadInfoMana
 
     private void prepare() {
         recyclerView.setAdapter(mDownloadListAdapter);
-        recyclerView.setLayoutManager(new LinearLayoutManager(getActivity(),LinearLayoutManager.VERTICAL,true));
+        recyclerView.setLayoutManager(new LinearLayoutManager(getActivity(),LinearLayoutManager.VERTICAL, false));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/fragment/DownloadsFragment.java
```diff
@@ -15,6 +15,7 @@
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.util.Log;
```
mTwTm: Redundant

app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -289,10 +296,14 @@ public int getItemCount() {
 
     @Override
     public void onQueryComplete(List downloadInfoList) {
-        mDownloadInfo.addAll(downloadInfoList);
+
```
mTwTm: Why do we replace addAll with foreach?

app/src/main/java/org/mozilla/focus/fragment/DownloadsFragment.java
```diff
@@ -76,7 +77,9 @@ public void onReceive(Context context, Intent intent) {
 
     @Override
     public void tryLoadMore() {
-        // TODO: 2017/9/26
+        if (!mDownloadListAdapter.isLoading()){
```
mTwTm: Could be improved: This will send unnecessary load mores whenever were scrolling down in the last page.

app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -45,22 +45,29 @@
...
         isOpening = true;
     }
 
+    public boolean isLoading(){
+        return isLoading;
+    }
+
     public void loadMore(){
-        DownloadInfoManager.getInstance().query(mItemCount,50,this);
+        DownloadInfoManager.getInstance().query(mItemCount,20,this);
```
mTwTm: move pageSize into a constant so we can easily change this.

mTwTm: In the meanwhile, I think 50 may be a better choice.

WeiWW: I afraid that querying  too mach at once may happen TransactionTooLargeException.

mTwTm:
Fixes #974

PR #1009, opened 2017-09-28
Only insert http or https scheme into history

walkingice:


walkingice:
this patch is supposed to fix #1001. According to its crash stack, it tried to open url `data:text/html;charset=utf-8;base64` and this patch won't let it happen.

app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -270,7 +270,9 @@ public void insertBrowsingHistory() {
             return;
         } else if (SupportUtils.BLANK_URL.equals(url)) {
             return;
-        } else if (!UrlUtils.isSupportedProtocol(url)) {
+        }
+
+        if (!UrlUtils.isHttpOrHttps(url)) {
```
mTwTm: This will block file:// is this expected behavior? Blocking it seems correct to me so r+ if this is not unexpected.

walkingice: Yes it is expected to block `file://` for now.

PR #1019, opened 2017-09-28
dev: only throws exception in dev-build

walkingice:
We are closing to release. To reduce crash log, now we only throws
Exception in dev-build to warn developers. And keep silent to other
people.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
@@ -153,10 +153,10 @@ public abstract class WebFragment extends LocaleAwareFragment {
             if (UrlUtils.isUrl(url)) {
                 this.pendingUrl = null; // clear pending url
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -153,10 +153,10 @@ public void loadUrl(@NonNull final String url) {
             if (UrlUtils.isUrl(url)) {
                 this.pendingUrl = null; // clear pending url
 
-                // in case of any unexpected path to here, to normalize URL in release build
-                final String target = AppConstants.isReleaseBuild() ? UrlUtils.normalize(url) : url;
+                // in case of any unexpected path to here, to normalize URL in beta/release build
+                final String target = AppConstants.isDevBuild() ? url: UrlUtils.normalize(url);
```
mTwTm: need space after url

walkingice: D'oh!

PR #1028, opened 2017-10-01
Don't update failing url when error page is being displayed

### LINKED ISSUE #1026: visit an unavailable page several times will put it to Top Sites (RC2)
STR:

1. Visit http://wesley.huang
2. tap on "TRY AGAIN" several time
3. Observe Top Sites

Build 1412 RC1
DuT: Redmi

mTwTm:
Fixes #1026 

Well this is assuming that failing url does not load correctly so it will always be redirected to our errorpage. An assumption that does not always hold but is fairly correct for the most of the time. Let's deal with the edge cases later.


hartsu:
LGTM

PR #1043, opened 2017-10-02
Improve loadMore performance

### LINKED ISSUE #1042: Loading more seems laggy
As title. 

**Build:** 1413

mTwTm:
Fixes #1042


mTwTm:
We shouldn't do this since we may cause TransactionTooLarge

Issue #1051, opened 2017-10-03
"OPEN" in the snack bar of files just downloaded can't work (RC3)


cssjh:
**STR**
1. Set save files to SD card
2. Download a file
3. When the download is finished, tap "OPEN" in the snack bar -> the file won't open

**DuT:** Samsung J7 Prime
**Build:** 1437 RC2

cssjh:
oops so awkward...:P
@chianger thanks for correcting!

dakaizou:
verified in build 1512

PR #1057, opened 2017-10-03
Send RELOCATE_FINISH broadcast after we updated db

mTwTm:
We're updating the rowId -> filePath mapping in replacePath()

This creates a race condition for queryByDownloadId -> updateByRowId vs Snackbar.show -> userClickAction

On Samsung, updateByRowId is usally slower.

Thus we should actually toast after we have finished the update.

BTW I should reuse the broadcast code... Let's fix it when we have time.


mTwTm:
History: In #1018 we have successfully wait until move is finished, but we have not wait until we have updated our DB, see the above for the description to the fix.

mTwTm:
Fixes #1051

PR #1065, opened 2017-10-05
fix #1054: if it is a new one, it should be added at first.

### LINKED ISSUE #1054: "Downloading" item is inserted into wrong position
In downloading, when open download panel open before the item inserted in, it will be add in the end. But open it again, it will be at first.

WeiWW:
as title


app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -86,7 +86,7 @@ public void updateItem(DownloadInfo downloadInfo){
         }
 
         if (index == -1){
-            mDownloadInfo.add(downloadInfo);
+            mDownloadInfo.add(0,downloadInfo);
```
mTwTm: Extra space needed :P

mTwTm:
Fixes #1054

Issue #837, opened 2017-09-19
[dogfooding] Press delete key on keyboard didn't remove autocomplete


Conjuror:
When searching, after inputing some words then press delete key on keyboard, the word will be displayed abnormally

dakaizou:
Possible related to #610

chianger:
QAWanted to have clearer STR so that we can have Fable support.

dakaizou:
STR:
1. Tap on search bar
2. Type in "moz"
3. Tap on device delete (the autocompleted text has been deleted, only user-typed text left)
4. Tap on "m" (now the "z" is replaced with "m")

Note: This only happens on devices using buffered input method (notice the underscore in the recording)

DuT: Oppo

Conjuror:
Note that this can be reproduce if you are typing with input method. For an example, use 注音 or 無蝦米, typing a word and tap on delete key. The buffered alphabet is still on the input field, until you type in another word/alphabet to replace it.

Suggest to P2 and put as low priority

Conjuror:
Remove QAwanted as STR is clear.

chianger:
Removing P2 items from sprint8.

mTwTm:
I'll take this one, thanks for your study @joseph-yar !

mTwTm:
Fixed in #1187

PR #1073, opened 2017-10-06
Fix input text is not aligned with IME when delete autocomplete text …

joseph-yar:
…under composing mode.

- #837


walkingice:
output

I got regression on this patch. When pressing del key, it removes suggestion and also offset cursor

PR #1093, opened 2017-10-11
fix #949 : change logo on about page.

### LINKED ISSUE #949: [Logo] Change the font style of "Firefox Rocket" in the home page

WeiWW:
as title.


app/src/main/res/layout/fragment_homescreen_item_title.xml
```diff
@@ -6,5 +6,5 @@
     android:id="@+id/home_fragment_title"
     android:layout_width="wrap_content"
     android:layout_height="wrap_content"
-    android:src="@drawable/logotype"
+    android:src="@drawable/logotype_black"
```
mTwTm: Let's just keep the xxhdpi and hdpi ones since we want to minimize the apk size. Let's also not change the code when we can avoid doing so. Let's add these resources with the same file name.

mTwTm:
@Ben-YPLai can you help review this after @WeiWW had fixed the things I've mentioned :)?

Ben-YPLai:
LGTM

Issue #1100, opened 2017-10-12
EXIT button


joechengla:

Jacklin-ux:
Just quick study. UC, UC mini, Opera mini all has "Exit" button on the menu

PR #1106, opened 2017-10-15
Add close button on main menu

haydarai:
Close https://github.com/mozilla-tw/Rocket/issues/1100

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -377,6 +377,10 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements Fragme
                 onPreferenceClicked();
                 TelemetryWrapper.clickMenuSettings();
                 break;
+            case R.id.menu_exit:
+                onExitClicked();
+                TelemetryWrapper.clickMenuExit();
+                break;
             case R.id.action_next:
             case R.id.action_loading:
             case R.id.action_share:
@@ -448,6 +452,11 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements Fragme
 
     private void onPreferenceClicked() {
         openPreferences();
+    }
+
+    private void onExitClicked() {
+        finish();
+        moveTaskToBack(true);
     }
 
     private void onDownloadClicked() {
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -456,7 +456,6 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements Fragme
 
     private void onExitClicked() {
         finish();
-        moveTaskToBack(true);
     }
 
     private void onDownloadClicked() {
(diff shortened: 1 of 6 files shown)


carol-mozilla:
Hello! please update the exit icon : https://drive.google.com/file/d/0B-9PIePlQZRlcjlvVWVmWm1kQzA/view?usp=sharing
Thanks!!

carol-mozilla:
Can I check the menu UI again once the new icon has been replaced? thank you!!

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -450,6 +454,11 @@ private void onPreferenceClicked() {
         openPreferences();
     }
 
+    private void onExitClicked() {
+        finish();
+        moveTaskToBack(true);
```
mTwTm: I don't think we need this after a finish?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -383,6 +383,10 @@ public static void clickMenuSettings() {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, Value.SETTINGS).queue();
     }
 
+    public static void clickMenuExit() {
+        TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, Value.EXIT).queue();
```
mTwTm: @idiotmax can you help make sure this event is ready?

app/src/main/res/drawable/menu_exit.xml
```diff
@@ -0,0 +1,12 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
mTwTm: We have another resource provided by our designer :) See the comment

app/src/main/res/values-in/strings.xml
```diff
@@ -73,6 +73,7 @@
   <string name="label_menu_block_ads">Blokir Iklan</string>
   <string name="label_menu_block_image">Blokir gambar</string>
   <string name="label_menu_settings">Setelan</string>
+  <string name="label_menu_exit">Keluar</string>
```
mTwTm: We have these localized through https://pontoon.mozilla.org/id/firefox-rocket/app.po/ Let's remove this and we may join the localization discussion in pontoon when the string lands there.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
haydarai: Okay, will refactor it later today 😄

app/src/main/res/values-in/strings.xml
haydarai: So for the time being it will used the English value, right?

mTwTm: Yeah simply leave the `strings.xml` in `values-in/` as-is should be good enough. (Only update the one in `values/` ) :)

haydarai:
@carol-mozilla do we have the xml version of the svg?

haydarai:
@mTwTm please do review it again one more time 😄

mTwTm:
@haydarai Sorry for switching back and forth, but after some short discussion we'd like to add the string-id again, can you help with this?

PR #1122, opened 2017-10-17
bug: to catch Exception in volume checking

### LINKED ISSUE #1088: java.lang.IllegalArgumentException
java.lang.IllegalArgumentException: 
 
  at android.os.Environment.isExternalStorageRemovable (Environment.java:1016)
 
  at org.mozilla.focus.utils.StorageUtils.getFirstRemovableMedia (StorageUtils.java:94)
 
  at org.mozilla.focus.utils.StorageUtils.getAppMediaDirOnRemovableStorage (StorageUtils.java:34)
 
  at org.mozilla.focus.utils.StorageUtils.getTargetDirOnRemovableStorageForDownloads (StorageUtils.java:62)
 
  at org.mozilla.focus.activity.MainActivity.asyncCheckStorage (MainActivity.java:775)
 
  at org.mozilla.focus.activity.MainActivity.access$300 (MainActivity.java:68)
 
  at org.mozilla.focus.activity.MainActivity$2.run (MainActivity.java:764)
 
  at java.lang.Thread.run (Thread.java:818)

walkingice:
to add try-catch in case of any unexpected situation which crash app.

to fix #1088

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/StorageUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/StorageUtils.java
@@ -90,9 +90,15 @@ public class StorageUtils {
     static File getFirstRemovableMedia(@NonNull Context ctx) {
         final File[] files = ctx.getExternalMediaDirs();
         for (final File file : files) {
-            // on some devices such as Oppo, it might return null
-            if ((file != null) && Environment.isExternalStorageRemovable(file)) {
-                return file;
+            try {
+                // on some devices such as Oppo, it might return null
+                if ((file != null) && Environment.isExternalStorageRemovable(file)) {
+                    return file;
+                }
+            } catch (IllegalArgumentException e) {
+                // Environment will invoke internal API to check volume,
+                // and throws an IllegalArgumentException if the target path has any problem
+                e.printStackTrace();
             }
         }
         return null;


app/src/main/java/org/mozilla/focus/utils/StorageUtils.java
```diff
@@ -91,8 +91,17 @@ static File getFirstRemovableMedia(@NonNull Context ctx) {
         final File[] files = ctx.getExternalMediaDirs();
         for (final File file : files) {
             // on some devices such as Oppo, it might return null
-            if ((file != null) && Environment.isExternalStorageRemovable(file)) {
-                return file;
+            if ((file != null)) {
+                // Environment will invoke internal API to check volume,
+                // and throws an IllegalArgumentException if the target path has any problem
+                try {
+                    if (Environment.isExternalStorageRemovable(file)) {
```
mTwTm: https://developer.android.com/reference/android/os/Environment.html#isExternalStorageRemovable() states it throws IllegalArgumentException if the path is not a valid storage device.

Let's not catch everything but just IllegalArgumentException since some thing/api may have changed and we want to know when it is broken.

app/src/main/java/org/mozilla/focus/utils/StorageUtils.java
```diff
@@ -91,8 +91,17 @@ static File getFirstRemovableMedia(@NonNull Context ctx) {
         final File[] files = ctx.getExternalMediaDirs();
         for (final File file : files) {
             // on some devices such as Oppo, it might return null
-            if ((file != null) && Environment.isExternalStorageRemovable(file)) {
-                return file;
+            if ((file != null)) {
```
mTwTm: Would it be simpler to just do:

try {
   if ((file != null) && Environment.isExternalStorageRemovable(file)) {
   }
}

walkingice:
updated to follow @mTwTm 's suggestion

PR #1134, opened 2017-10-18
Sort by ID rather then date added when query latest image added

### LINKED ISSUE #1133: [Smoke Test] Download triggers screenshot onboarding if moved to sdcard and last entry is screenshot
STR:
1. Clear data or reinstall rocket
2. Grant storage permission (download any picture)
3. Tap on system Home
4. Tap on system screenshot
5. Go back to rocket
6. Download any picture

Build: 1541
DuT: Samsung

mTwTm:
Fixes #1133

### DIFF
--- a/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotObserver.java
+++ b/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotObserver.java
@@ -11,6 +11,10 @@ import org.mozilla.focus.utils.ThreadUtils;
 
 public class ScreenshotObserver {
     public Context context;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotObserver.java
```diff
@@ -78,7 +78,7 @@ public void onMediaChange(final Uri uri) {
             @Override
             public void run() {
                 // Find the most recent image added to the MediaStore and see if it's a screenshot.
```
walkingice: I think we should update the comment from 'added' to 'modified'

Issue #1136, opened 2017-10-18
Won't open Rocket app from task manager after exiting app by pressing Exit button


Jacklin-ux:
After pressing Exit button on the menu to close Rocket, users open system task manager then select Rocket to open. Current behavior is swipe out Rocket and won't open it, but UX expected behavior is open Rocket.

haydarai:
Since the `minSdkVersion` is 21, we can use this https://developer.android.com/reference/android/app/ActivityManager.AppTask.html#finishAndRemoveTask() to exit the app and remove it from recent apps list. I'll make the changes later today.

dakaizou:
@Jacklin-ux Can you specify the device you're using. QA's devices seemed working fine

mTwTm:
My Sony XA can reproduce. I think I've seen it on pixel too.

dakaizou:
It's happening on device with Android 7.0+

Jacklin-ux:
I saw it on my Pixel phone

walkingice:
We have `<activity-alias>` for MainActivity, and set `android:exported="false"` for **org.mozilla.focus.activity.MainActivity**.  Once we switch apps view Recent-apps, system send intent to **org.mozilla.focus.activity.MainActivity** which is not exported.

```
10-23 15:15:20.465   904  1576 I ActivityManager: START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10300000 cmp=org.mozilla.rocket.debug.jchu$
org.mozilla.focus.activity.MainActivity} from uid 10032
10-23 15:15:20.465   904  1576 W ActivityManager: Permission Denial: starting Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10300000 cmp=or$
.mozilla.rocket.debug.jchu/org.mozilla.focus.activity.MainActivity } from null (pid=-1, uid=10032) not exported from uid 10164
10-23 15:15:20.469  1419  1530 E SystemServicesProxy: Could not start Firefox Rocket Dev.
10-23 15:15:20.469  1419  1530 E SystemServicesProxy: java.lang.SecurityException: Permission Denial: starting Intent { act=android.intent.action.MAIN cat=[android.intent.categor$
.LAUNCHER] flg=0x10300000 cmp=org.mozilla.rocket.debug.jchu/org.mozilla.focus.activity.MainActivity } from null (pid=-1, uid=10032) not exported from uid 10164
[...]

PR #1142, opened 2017-10-18
Refactor exit button to exit app and remove Rocket app from task manager

haydarai:
close https://github.com/mozilla-tw/Rocket/issues/1136


mTwTm:
@haydarai We're leaving the app in the recent task intentionally. However there is a bug which we're accidentally finishing the app when the Activity is recycled. (and is what that bug is about) If you're interested in fixing this bug, feel free to consult @walkingice or me.

PR #1143, opened 2017-10-19
Add feedback page into browsing stack.

### LINKED ISSUE #1116: Close user feedback page then back to the previous webpage via system back key
Expected behavior:
Open the survey via "Settings-Give feedback-No, send feedback",  press system back key then back to the previous webpage instead of closing app. 
(Put feedback page into browsing stack)

joseph-yar:
- Closes #1116


joseph-yar:
There are two entry point for launching share app dialog, “Settings->Give feedback” and “MainActivity->App initiation check”. I think onActivityResult() is ok for the first one, but not 100% appropriate for the second one. The other reason for creating an open url method in IntentUtils is we continually got some requests that need to add the url into browsing stack. Maybe we can utilize this mechanism for them and future request.

PR #1144, opened 2017-10-19
fix #829 : set to landscape mode onEnterFullScreen, and set to portra…

### LINKED ISSUE #829: [dogfooding] Press expand icon on youtube video, suggest to play full screen video
as title

WeiWW:
…it mode onExitFullScreen.


walkingice:
discussed with @mTwTm , since we also set orientation config in method `switchToImmersiveMode()`, thing might be more complicated than we imagined.

PR #1148, opened 2017-10-19
Workaround for #907.

### LINKED ISSUE #907: Refresh is only on error pages rather than the page if pages fails before loaded
STR:

1. Turn off network
2. Load any page
3. One should receive the error page
4. Turn on network and refresh
5. One is still seeing the error page (Expected: the page should be loaded.)

Comparison:
If in step 4 "Try again" button is clicked, page will load successfully
If we first load successfully before step 1, then refresh in step 2, page will load successfully in step 5.

hartsu:
Closes #907

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
+++ b/app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
@@ -160,6 +160,14 @@ public class WebkitView extends NestedWebView implements IWebView, SharedPrefere
         client.notifyCurrentURL(url);
     }
 
+    public void reload() {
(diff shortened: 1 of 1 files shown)


walkingice:
Looks like we need more work on fixing this issue. It won't reload in below situation

1. Launch Rocket with connected network
1. Open any web page, and it would load correctly
1. turn off WiFi, reload -> now we got error page
1. turn on WiFi, reload -> got white screen
1. reload again -> even URL becomes internal-error-url

walkingice:
It seems the previous comment is only reproduce-able on Android 8, and that is not our target so far. I think we can land this patch.

PR #1152, opened 2017-10-20
fix #842: Can't open keyboard after entering & leaving full screen on…

### LINKED ISSUE #842: Can't open keyboard after entering & leaving full screen on YouTube
**STR**
_for Android 7.0+_
1. Open a YouTube video
2. Enter full screen
3. Leave full screen
4. Tap on YouTube search button -> keyboard won't appear

**DuT**: Pixel, Zenfone 3
**Build**: 1105

Ben-YPLai:
… YouTube

WebView gets focus, but unable to open the keyboard after exit Fullscreen for Android 7.0+
I guess some component in WebView might lock focus
So when user touches the input text box on Webview, it will not trigger to open the keyboard
It may be a WebView bug.
The workaround is clearing WebView focus
The WebView will be normal when it gets focus again.
If android change behavior after, can remove this.

close #842

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -472,6 +472,18 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
                     fullscreenCallback.fullScreenExited();
                     fullscreenCallback = null;
                 }
+
+                // WebView gets focus, but unable to open the keyboard after exit Fullscreen for Android 7.0+
+                // We guess some component in WebView might lock focus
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -472,6 +472,13 @@ public void onExitFullScreen() {
                     fullscreenCallback.fullScreenExited();
                     fullscreenCallback = null;
                 }
+
+                // WebView unable to open keyboard after exit Fullscreen for Android 7.0+
```
idiotmax: Please update comment to match your experiment and findings.

PR #1154, opened 2017-10-20
Improve download image.

WeiWW:
If we guess file name from image url, and then we get the file name end with ".bin", we will get HEAD at first in order to get Content-type(MIME type).

### DIFF
--- a/app/src/main/java/org/mozilla/focus/download/GetImgHeaderTask.java
+++ b/app/src/main/java/org/mozilla/focus/download/GetImgHeaderTask.java
@@ -1,0 +1,60 @@
+package org.mozilla.focus.download;
+
+import android.os.AsyncTask;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Created by anlin on 20/10/2017.
+ */
+
+public class GetImgHeaderTask extends AsyncTask<String, Void, String> {
+
+    public interface Callback {
+        void setMIMEType(String mimeType);
+    }
+
+    public GetImgHeaderTask.Callback callback;
+
+    public void setCallback(Callback callback) {
+        this.callback = callback;
+    }
+
+    @Override
+    protected String doInBackground(String... params) {
+        HttpURLConnection connection = null;
+        String contentType = "";
+        int responseCode = 0;
+        try {
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/download/GetImgHeaderTask.java
```diff
@@ -0,0 +1,66 @@
...
+        try {
+
+            connection = (HttpURLConnection) new URL(params[0]).openConnection();
+            connection.setRequestMethod("HEAD");
+            contentType = connection.getContentType();
+            responseCode = connection.getResponseCode();
+
+            connection.disconnect();
+
+        } catch (ProtocolException e) {
```
mTwTm: try
```
        } catch (ProtocolException | MalformedURLException e) {
            e.printStackTrace();
        } catch (IOException e) {
            e.printStackTrace();
```
or 
```
        } catch (ProtocolException | MalformedURLException | IOException e) {
            e.printStackTrace();
```

app/src/main/java/org/mozilla/focus/download/GetImgHeaderTask.java
```diff
@@ -0,0 +1,66 @@
...
+        } finally {
+            if (connection != null) {
+                connection.disconnect();
+            }
+        }
+
+        if (responseCode == HttpURLConnection.HTTP_OK) {
+            return contentType;
+        } else {
+            return "";
```
mTwTm: return null should be better? So we shall align the behavior to the old behavior

app/src/main/java/org/mozilla/focus/menu/WebContextMenu.java
```diff
@@ -118,8 +120,22 @@ public boolean onNavigationItemSelected(@NonNull MenuItem item) {
                         return true;
                     }
                     case R.id.menu_image_save: {
-                        final Download download = new Download(hitTarget.imageURL, null, null, null, -1, true);
-                        callback.onDownloadStart(download);
+                        if (URLUtil.guessFileName(hitTarget.imageURL, null, null).endsWith(".bin")) {
```
mTwTm: Let's don't hardcode ".bin" here, Declare:
```
public static final DEFAULT_DOWNLOAD_EXTENSION = ".bin";
```
in `URLUtil` and change them respectively

PR #1183, opened 2017-10-26
The crash happened when data set size is 1 after add item. So in this…

WeiWW:
The crash will happen when data set size is 1 after add item. Because we define item count is 1 and mDownloadInfo is empty that means nothing and show empty view. So use notifyDataSetChanged() instead of notifyItemInserted() when data size is 1 after add item.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
@@ -158,7 +158,15 @@ public class DownloadListAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
 
         if (index == -1){
             mDownloadInfo.add(downloadInfo);
-            this.notifyItemInserted(mDownloadInfo.size()-1);
+            //The crash will happen when data set size is 1 after add item.
+            //Because we define item count is 1 and mDownloadInfo is empty that means nothing and show empty view.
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -158,7 +158,12 @@ private void addItem(DownloadInfo downloadInfo){
 
         if (index == -1){
             mDownloadInfo.add(downloadInfo);
-            this.notifyItemInserted(mDownloadInfo.size()-1);
+            if (mDownloadInfo.size() > 1){
```
mTwTm: Are we giving the incorrect index?
Simply fix with
`this.notifyItemInserted(mDownloadInfo.size());` ?

app/src/main/java/org/mozilla/focus/widget/DownloadListAdapter.java
```diff
@@ -158,7 +158,12 @@ private void addItem(DownloadInfo downloadInfo){
...
+                this.notifyItemInserted(mDownloadInfo.size()-1);
+            }else{
+                this.notifyDataSetChanged();
```
mTwTm: Make a comment around we cannot use notifyItemInserted since we have empty view

PR #1187, opened 2017-10-26
Remove setComposingRegion() in removeAutocompleteOnComposing

mTwTm:
We assumes IME will sync its state itself after we call
finishComposingText and setComposingRegion, this is not the
case for Google Zhuying and Japanese, remove them to avoid
making the experience even worse.


joseph-yar:
LGTM!

Issue #40, opened 2017-07-26
Gesture: Pull to refresh


chianger:

chianger:
this was triaged to be P2 feature

PR #1193, opened 2017-10-31
Add pull to refresh (#40)

cnevinc:


cnevinc:
There are some issues for pull to refresh now.
See 
https://github.com/mozilla-mobile/focus-android/issues/1776
https://github.com/mozilla-mobile/focus-android/issues/1807

two good examples are 
http://www.independent.co.uk/
http://maps.google.com

I'll investigate more.

Issue #1217, opened 2017-11-05
cannot directly open it via the pop up notification if the file name conatins " " (space)


vanceyen:
STR:
1.  Download a MP3 file from http://www.planetlagu.site/
2. When download complete, a notification will popup in the bottom of the screen
3. Select “Open” in the notification. 

Expected result:
Something will happen to allow user listen to the MP3 she just downloads

Actual result:
Nothing happen. But works fine if user tries to play the download MP3 from download list

DUT: Any device can reproduce this issue
Build: 1.0 - 1692

chianger:
user has workaround to go to 'download' panel and able to play. 
need to look into this as it seems to be a regression.

chianger:
close as #1225 is merged.

PR #1225, opened 2017-11-06
Remove Media scan when downloaded snackbar open fired.

mTwTm:
We now have FileProvider, this is not needed now. In the meanwhile
there's a bug here which path is provided but uri is wanted. This
raises some problems such as unencoded space is not a valid Uri.


mTwTm:
Fixes #1217

PR #1246, opened 2017-11-17
Enable checkstyle

mTwTm:
Enable checkstyle and pmd with BuddyBuild preBuild scripts.


mTwTm:
Hmm I'm seeing this in the last build log:

129 360 PMD rule violations were found. See the report at: file:///tmp/sandbox/workspace/app/build/reports/pmd/pmd.html

but not seeing it locally. Maybe I should just add checkstyle first

PR #1254, opened 2017-11-24
Update firstLoadingUrlAfterResumed when url changes too

### LINKED ISSUE #1192: Won't see refresh button on the menu when back to previous site without showing loading bar
Reproduce,
1. Go to Mozilla.org page from Home
2. Press home button on the bottom bar then open Home
3. Press system back key to back to Mozilla.org page (but there is no loading bar when back to Mozilla.org)

Expected,
Refresh button will be on the menu instead of X button

Actual,
X button is on the menu and will not be changed to Refresh button

mTwTm:
The previous implementation is ignoring cases when user browse
via url or when user have navigated to another page.
Hopefully this should be fixed with this commit.

Fixes #1192


mTwTm:
@walkingice Sounds like a good idea. I'm a little wearied by the fact that this implementation would strengthen the binding between `BrowserFragment` and `IWebView.Callback` but let's fix this with multi-tab

PR #1255, opened 2017-11-24
ui: display minimal progress value in loading page

### LINKED ISSUE #1232: sometimes rocket will not show loading bar when it has started loading

walkingice:
If we are loading page but the first reponse comes lately, the
progressbar might be 0 and user is not aware of that we are loading.

For perceive performance, to display 10% as min-value. It is same
as Fennec which pick 10 for LOAD_PROGRESS_INIT.

to fix #1232


mTwTm:
LGTM

Issue #1275, opened 2017-12-03
Highlighted text in search suggestion is case sensitive


dakaizou:
STR:
 1. Change default search engine to Google
 2. Tap the search button
 3. Type in "ama"
 4. __The "ama" is highlighted__
 5. Delete all text
 6. Type in "Ama"
 7. __The "ama" is not highlighted__

Expected Result:
 - "ama" is highlighted in either situation
 - Chrome "ama" !["ama"](https://lh3.googleusercontent.com/-zb0wxUQLU3w/WiQvReCvs4I/AAAAAAAAB20/oOv1bFsUYDE2kJQAD_lt7UmRjr-Cm7UGACL0BGAYYCw/h1920/5475711844103102992%253Faccount_id%253D1)
 - Chrome "Ama" !["Ama"](https://lh3.googleusercontent.com/-AyIpIZXMza0/WiQvQhI1a5I/AAAAAAAAB2g/nHSfgJBD94wUX9yz6uwVeYJUUHvLd62HQCL0BGAYYCw/h1920/3696967860077138009%253Faccount_id%253D1)

 Actual Result:
 - "ama" is only highlighted if it's a exact match
 - "ama" !["ama"](https://lh3.googleusercontent.com/-dojCc2ZprP0/WiQwSbHqtAI/AAAAAAAAB3g/uOSWm3kfoQofM6AWG6EepM-E63WTp0dUQCL0BGAYYCw/h1920/4804086780313795967%253Faccount_id%253D1)
 - "Ama" !["Ama"](https://lh3.googleusercontent.com/-_5CAR2h2eFc/WiQwRyTYQPI/AAAAAAAAB3U/DAUKHGICin8qvnyyKWwb4mm1iNoCs9kUwCL0BGAYYCw/h1920/3459428500421394546%253Faccount_id%253D1)

Reproduction rate: 5/5
Rocket version: 1.0.1 Build 1777

Device name: Samsung S7 edge
Android version: 7.0

Device name: Redmi 4A
Android version: 6.0

PR #1280, opened 2017-12-04
fix issue #1275

ariefbayu:
Hi Guys, I tried to get low hanging fruit to get a better understanding of the code structures. Here's my take. Please confirm if it's good or need further modification.

below are screenshots of my test...

photo_2017-12-04_16-18-36

photo_2017-12-04_16-18-38

### DIFF
--- a/app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.java
+++ b/app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.java
@@ -245,7 +245,7 @@ public class UrlInputFragment extends Fragment implements UrlInputContract.View,
         for (int i = 0; i < texts.size(); i++) {
             final TextView item = (TextView) View.inflate(getContext(), R.layout.tag_text, null);
             final String str = texts.get(i).toString();
-            final int idx = str.indexOf(searchKey);
+            final int idx = str.toLowerCase().indexOf(searchKey);
             if (idx != -1) {
                 SpannableStringBuilder builder = new SpannableStringBuilder(texts.get(i));
                 builder.setSpan(new android.text.style.StyleSpan(android.graphics.Typeface.BOLD),


mTwTm:
Thanks for the pr!!

app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.java
```diff
@@ -241,7 +241,7 @@ public void setSuggestions(@Nullable List<CharSequence> texts) {
             return;
         }
 
-        final String searchKey = urlView.getOriginalText().trim();
+        final String searchKey = urlView.getOriginalText().trim().toLowerCase();
```
mTwTm: Looks like a good solution to me. I've got some other things in my mind, though:

A foreseeable situation I can see is this will not work when the user is using Turkish locale. which has a special lowercase of I : ı https://en.wikipedia.org/wiki/Dotted_and_dotless_I

so I would recommend:
final String searchKey = urlView.getOriginalText().trim().toLowerCase(Locale.US);

Another situation is, I believe some search suggestions are provided in capital: I haven't found an example in Google's suggestions but you can try "CD" for duckduckgo, this solution also break under suggestions like this. I'm not sure if lowering every suggestion sounds like a good idea. I'll leave it to the reviewer to decide :P @walkingice can you help review this?

mTwTm: Looks like a good solution to me. I've got some other things in my mind, though:

A foreseeable situation I can see is this will not work when the user is using Turkish locale. which has a special lowercase of I : ı https://en.wikipedia.org/wiki/Dotted_and_dotless_I

so I would recommend:
final String searchKey = urlView.getOriginalText().trim().toLowerCase(Locale.US);

Another situation is, I believe some search suggestions are provided in capital: I haven't found an example in Google's suggestions but you can try "CD" for duckduckgo, this solution also break under suggestions like this. I'm not sure if lowering every suggestion sounds like a good idea. I'll leave it to the reviewer to decide :P @walkingice can you help review this?

ariefbayu: Yeah, I'm not that knowledgeable in regard to locale / localization, so, your input it highly appreciated. I'll wait what the reviewer said and will follow up accordingly...

mTwTm:
@walkingice is blocked by other tasks, I'll takeover the review

app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.java
mTwTm: Let's first start with lowering the both since wikipedia search suggestions are always capitalized. I'll take a look at if there is any performance issue here. Since we're toLowerCase()ing both strings, I guess the locale would be irrelevant.

ariefbayu: I'm not sure by "lowering the both", can you elaborate?

mTwTm: In:

`str.indexOf(searchKey);`

do 

`str.toLowerCase().indexOf(searchKey);` instead

mTwTm:
Hi! Sorry for the late response, the whole team is currently traveling to join the Mozilla meetup. I've just tried the solution and the solution seems great including performance-wise.

In our workflow we usually just pushes the patches (the second commit) on the branch (say, ariefbayu:master and github will automatically reflect the change, you can see that's true since 
fc52c34 is included in this pull request.) And merge the commits if needed after the review. However I agree that it's usually not a good idea to do things on master. Let's do the work on master this time but I'd say life will be easier for you if you merge from another branch.

One last thing: can you squash the commits into one commit since I believe the two toLower()s are inseparable?

Let me know if you need any help!

ariefbayu:
I've redo the commit and pushed it to a branch in my cloned repo. Should I create new PR?

mTwTm:
Seems like you have deleted the original cloned repo (and cloned again)? With that then yes I guess we will need a new PR. I was proposing to reuse the original fork's master (pushing to the branch also reflects on this original PR). I'm proposing so because with that way we can keep our conversation in one place in case any one is interested :)

I took a look on your branch and think it's ready to go, let's do this :)

ariefbayu:
Haha, because I'm not sure what to do and how to `squash the commits into one commit`. I guess I'll learn that in another issues that I might tackle.

mTwTm:
@ariefbayu Give this a try :) https://shinglyu.github.io/web/2016/11/08/servo-rebase-and-squash-guide.html

mTwTm:
Closing this PR, continuing at #1283

PR #1290, opened 2018-01-03
new permission model

### LINKED ISSUE #1260: handle situations where one or many permissions have been denied by the user
The app must be prepared to handle situations where one or many permissions have been denied by the user. Educate the user on why the permission is needed and allow the user to trigger the prompt again either immediately or when they have the related action.  
For more info:
https://material.io/guidelines/patterns/permissions.html#permissions-denied-permissions

### LINKED ISSUE #1261: In the event of a user selecting “Don’t ask again” for a permission tied to a particular function, the app should let the user know why they cannot proceed with the desired function
- In the event of a user selecting “Don’t ask again” for a permission tied to a particular function, the app should let the user know why they cannot proceed with the desired function. Instruct the user on what actions they must take to re-enable the desired function (e.g direct user to app settings and what to do at the app settings view).  
For more info: https://material.io/guidelines/patterns/permissions.html#permissions-denied-permissions

mTwTm:
Fixes #1260 
Fixes #1261

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -138,6 +138,11 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements Fragme
             }
 
             @Override
+            public void doActionNoPermission(String permission, int actionId, Parcelable params) {
+                // Do nothing
+            }
+
+            @Override
             public int getDoNotAskAgainDialogString(int actionId) {
                 if (actionId == ACTION_CAPTURE ) {
                     return R.string.permission_dialog_msg_storage;
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -15,6 +15,7 @@ import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.os.Bundle;
 import android.os.Handler;
+import android.os.Parcelable;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


mTwTm:
This is a draft of the architecture to be used to solve the problem. Only partial of the fix is implemented for now, please do not merge this pr.

mTwTm:
@idiotmax Can you help take a look at this solution before I move on and implement it into full?

mTwTm:
This branch is ready to be reviewed.

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -0,0 +1,152 @@
...
+    private String permission;
+    private int actionId;
+    private Serializable params;
+
+    public PermissionHandler(PermissionHandle permissionHandle) {
+        this.permissionHandle = permissionHandle;
+    }
+
+    public void tryAction(final Activity activity, final String permission, final int actionId, final Serializable params) {
+        if (PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(activity, permission)) {
```
idiotmax: Try to reduce if/else depth level, maybe "early return" would be enough.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -101,6 +102,60 @@ protected void onCreate(Bundle savedInstanceState) {
         initBroadcastReceivers();
 
         mediator = new MainMediator(this);
+        permissionHandler = new PermissionHandler(new PermissionHandle() {
+            @Override
+            public void doAction(String permission, int actionId, int triggerType, Serializable params) {
+                if (actionId == ACTION_CAPTURE) {
+                    BrowserFragment browserFragment = getBrowserFragment();
+                    switch (triggerType) {
```
idiotmax: Maybe check browserFragment null and early return before entering switch case?

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -0,0 +1,152 @@
...
+            if (PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(activity, permission)) {
+                permissionHandle.doAction(permission, actionId, PermissionHandle.TRIGGER_SETTING, params);
+            }
+        }
+        clearAction();
+    }
+
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        if (requestCode == actionId) {
+            if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
```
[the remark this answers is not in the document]
mTwTm: We don't have that use case yet so I'm thinking of extend that when we need to. This model does restrict the use case of only requesting one at a time.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
mTwTm: I don't assume every new action in the future will need a BrowserFragment.

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
mTwTm: Done!

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -0,0 +1,152 @@
...
+            // We do not have the permission to write to the external storage. Request the permission and start the
+            // capture from onRequestPermissionsResult().
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                setAction(permission, actionId, params);
+                // First permission ask, Never ask me again or not able to grand the permission
+                if (!isFirstTimeAsking(activity, permission) && !ActivityCompat.shouldShowRequestPermissionRationale(activity, Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
+                    // TODO: 1/3/18
+                    // This will also be shown when the device is not able to grant the permission
+                    // We might want to deal with this at some point?
+                    AlertDialog.Builder builder = new AlertDialog.Builder(activity);
```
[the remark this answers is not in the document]
mTwTm: Done!

PR #1294, opened 2018-01-05
Add webview version to about page & telemetry, closes #1288

### LINKED ISSUE #1288: add telemetry to get the webview version

idiotmax:


app/src/main/java/org/mozilla/focus/utils/DebugUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import android.text.TextUtils;
+import android.webkit.WebView;
+
+public final class DebugUtils {
+
+    private DebugUtils() {
+
+    }
+
+    public static final String loadWebViewVersion(Context context) {
```
mTwTm: There is a WebSettings.getDefaultUserAgent(context) that may be handy.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -188,6 +189,7 @@ public static void init(Context context) {
             final boolean telemetryEnabled = isTelemetryEnabled(context);
 
             updateDefaultBrowserStatus(context);
+            updatePrefValue(context, resources.getString(R.string.pref_key_webview_version), DebugUtils.loadWebViewVersion(context));
```
mTwTm: If we really care about performance, will it make sense to delay this update? We might benefit from using something like AlarmManager or any other scheduling mechanism. This will delay the update to the next onCreate() but I think it should be fine.

mTwTm: However, I feel like to correctly evaluate the severity of performance issues, we'll need metrics to measure our performance which we don't have yet.

app/src/main/java/org/mozilla/focus/utils/DebugUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import android.text.TextUtils;
+import android.webkit.WebView;
+
+public final class DebugUtils {
+
+    private DebugUtils() {
+
+    }
+
+    public static String loadWebViewVersion(Context context) {
```
mTwTm: I'm curious of the reason why WebSettings.getDefaultUserAgent(context) is not used directly here.

idiotmax: I think you're right. Using getDefaultUserAgent will make this more simple.

idiotmax: There's a UnitTest fail in WebSettings.getDefaultUserAgent. I will rollback to use the original plan for now.

liuche:
@idiotmax thanks, can you fill out this form, and include it in a comment? Also please include the values that could be collected, specifically here: https://github.com/mozilla-tw/Rocket/pull/1294/files#diff-55c9e4450f1e97f3978fbd05a746cde7R35

Form: https://github.com/mozilla/data-review/blob/master/request.md

Issue #957, opened 2017-09-26
top site favicon disappears randomly 


joechengla:
e.g.
Visit Twitter and wait for loading to finish
then press home and observe that Twitter has favicon
Now I visit Instagram from one of the top site choices
Wait for Instagram to finish loading and press home
Now observe that Twitter favicon disappears

Conjuror:
may related to #915

chianger:
can we set this P1.5?

Issue #586, opened 2017-09-05
[dogfooding] Top site ranking mechanism should be refined


Conjuror:
Top site ranking doesn't seem to work (all Yahoo on Joe's screen)

Build 724

mTwTm:
We're tripling the count since we insert in onReceivedIcon and onPageFinished

mTwTm:
onPageFinished is also fired twice or 4 times as much as onPageStarted

mTwTm:
09-06 15:05:48.368 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 260879366 - onPageStarted, url: https://m.youtube.com/
09-06 15:05:48.368 9254-9394/org.mozilla.zerda.debug.tyu E/libEGL: validate_display:99 error 3008 (EGL_BAD_DISPLAY)
09-06 15:05:48.843 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 260879366 - onPageFinished, url: https://m.youtube.com/
09-06 15:05:49.008 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 260879366 - onPageFinished, url: https://m.youtube.com/
09-06 15:05:50.760 1336-1352/? E/ReloadingLock: Finalizing LOCKED Token[shortcuts 2017-09-06 15:05:37.138]
09-06 15:05:55.144 639-673/? E/ANDR-PERF-OPTSHANDLER: Warning: Resource [2, 0] not supported for core 1. Instead use resource for core 0
09-06 15:05:55.144 639-673/? E/ANDR-PERF-RESOURCEQS: Failed to apply optimization [2, 2, 0]
09-06 15:05:55.198 9254-9254/org.mozilla.zerda.debug.tyu E/RecyclerView: No adapter attached; skipping layout
09-06 15:05:57.069 9254-9254/org.mozilla.zerda.debug.tyu E/RecyclerView: No adapter attached; skipping layout
09-06 15:05:57.322 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 250587576 - onPageFinished, url: https://m.youtube.com/
09-06 15:05:57.330 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 250587576 - onPageFinished, url: https://m.youtube.com/
09-06 15:05:57.344 9254-9254/org.mozilla.zerda.debug.tyu E/mTwTm: 250587576 - onPageStarted, url: https://m.youtube.com/
[...]

Issue #386, opened 2017-08-23
The title of the same top site tiles are changing


mTwTm:
STR:

Click on top site A, click home, click on top site b, click home, for the same tile, sometimes url is shown while sometimes page title is shown.

See the Mozilla tile in following screenshots:

![screenshot_20170823-130712](https://user-images.githubusercontent.com/6755074/29599735-5105ec4c-8804-11e7-8650-4c8667d5efae.png)
![screenshot_20170823-130725](https://user-images.githubusercontent.com/6755074/29599736-510ccf80-8804-11e7-8c02-68dfae02c7bf.png)

mTwTm:
STR: 
1. Clicking a page with a title again and again to make it a top site (It will be labeled with its title, such as Yotubue if you're trying with yotube.com)
2. Click another top site, the youtube topsite label becomes http://m.youtu....

dakaizou:
verified in build 790

Issue #334, opened 2017-08-21
[Smoke Test] Progress bar does not work functionally


Conjuror:
Build 439

DuT: Advan

Attempt to load Amazon.com, no progress bar appears on the screen

dakaizou:
improved in build 623, but no fading out
DuT: Samsung J7 Prime, Advan

Conjuror:
As @dakaizou mentioned, it may related to #277 and worth to take to a look.

mTwTm:
Some random log suggesting we probably shouldn't trust onPageStarted and onPageFinished for this purpose.

09-15 17:55:52.039 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: 10%
09-15 17:55:52.100 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: 10%
09-15 17:55:52.102 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted : http://gmail.com/
09-15 17:55:52.102 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted
09-15 17:55:52.130 11065-12123/org.mozilla.rocket.debug.null E/mTwTm: Loading : http://gmail.com/
09-15 17:55:52.319 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted : https://www.google.com/gmail/
09-15 17:55:52.319 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted
09-15 17:55:52.935 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted : https://mail.google.com/mail/
09-15 17:55:52.935 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted
09-15 17:55:53.208 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted : https://accounts.google.com/ServiceLogin?service=mail&passive=true&continue=https://mail.google.com/mail/&ss=1&scc=1&ltmpl=ecobx&nui=5&btmpl=mobile&emr=1&osid=1#
09-15 17:55:53.208 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: onPageStarted
09-15 17:55:53.396 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: 20%
09-15 17:55:53.598 11065-11065/org.mozilla.rocket.debug.null E/mTwTm: 25%
[...]

PR #1296, opened 2018-01-08
Revert work-around patches for webview(62~before), closed #1295

### LINKED ISSUE #1295: The loading progress bar stuck at 10%
STR: 
 1. Launch Rocket
 2. Navigate to any website via Url. (ex: mozilla.org)
 3. Press home button 
 3. Press system back key 

Expected Result:
 - User can back to the previous website.
 - The loading progress bar should not stuck 

Actual Result:
 - User can back to the previous website.
 - The loading progress bar stuck at 10%

Device: Sony Z3C
OS version: Android 6.0.1
WebView version: 62.0.3202.73
Rocket version: 1786, 1792

Device: Sony Z3C
OS version: Android 6.0.1
WebView version: 63.0.3239.1111
Rocket version: 1786, 1792

idiotmax:
Revert patches below

79d7f54
f46b260
c6f3b88

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -100,8 +100,6 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
     private final static int NONE = -1;
     private int systemVisibility = NONE;
 
-    private String firstLoadingUrlAfterResumed = null;
-
     public static BrowserFragment create(@NonNull String url) {
         Bundle arguments = new Bundle();
         arguments.putString(ARGUMENT_URL, url);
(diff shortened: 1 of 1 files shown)


mTwTm:
I have not tried but from tracking the issues this may regress: #957, #334 (on webview pre63), #586,  and could regress #386

mTwTm:
It is regressing #386 and #957 on webview 55, on webview 63 the symptom is sometimes title is replaced by another site's title (Not sure if this happens before this patch, though)
#334 seems good on 63!
#586 should be regressing on both 62 and 63 too, but is not very obvious by our current design.

I think we can give this patch a try once we're sure the pre-62 behavior is acceptable, or another way to do this is to branch testing 62/63 in these callbacks.

PR #1303, opened 2018-01-10
Max/task/build upgrades

idiotmax:


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -268,7 +269,7 @@ private void initViews() {
     }
 
     private void setUpMenu() {
-        final View sheet = getLayoutInflater().inflate(R.layout.bottom_sheet_main_menu, null);
+        final View sheet = getLayoutInflater().inflate(R.layout.bottom_sheet_main_menu, (ViewGroup) null);
```
mTwTm: This seems unnecessary.

app/src/main/java/org/mozilla/focus/customtabs/CustomTabsService.java
```diff
@@ -56,6 +56,11 @@ public int postMessage(ICustomTabsCallback iCustomTabsCallback, String s,
                                    Bundle bundle) throws RemoteException {
                 return 0;
             }
+
+            @Override
+            public boolean validateRelationship(ICustomTabsCallback iCustomTabsCallback, int i, Uri uri, Bundle bundle) throws RemoteException {
```
mTwTm: I have no idea what this is @walkingice can you help take a look at this part?

app/src/main/res/layout/toolbar.xml
```diff
@@ -43,8 +43,6 @@
         android:ellipsize="none"
         android:gravity="center_vertical"
         android:lineSpacingExtra="2sp"
-        android:lines="1"
```
mTwTm: I think we're favoring the one deleted here, remove ellipsize instead.

app/src/main/res/raw/third_party_licenses
```diff
@@ -1,11 +1,3 @@
-http://www.opensource.org/licenses/bsd-licensehttp://www.apache.org/licenses/LICENSE-2.0.txt
```
mTwTm: We're removing some items, is this intended?

mTwTm:
Let's rebase (I tried and seems good) and fix the unit test and we're ready to merge. (See above for some reminders.)

app/src/main/java/org/mozilla/focus/customtabs/CustomTabsService.java
walkingice: We can refer this API: https://developer.android.com/reference/android/support/customtabs/CustomTabsSession.html#validateRelationship(int,%20android.net.Uri,%20android.os.Bundle)

Since we don't support CustomTabs yet, return `false` looks good to me.

app/src/test/java/org/mozilla/focus/utils/SupportUtilsTest.java
```diff
@@ -30,12 +32,20 @@ public void cleanup() {
      */
```
mTwTm: Remove this comment, or state that there seems to be a Robolectric behavior change?

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
idiotmax: There's a lint warning about inflate should pass a view parent, not null. In our case, we inflate the view for a Dialog to setView, there's no way we can have the reference of it. This bypass the lint warning. 
However, this is only a warning not a fatal error of lint. maybe I should just skip this block.

mTwTm: It's a warning and from code quality perspective I don't think adding this is enhancing it. (Actually a new warning popups: Unnecessary cast) so I think the better approach would be ignoring this lint case by case when we're 100% sure we don't have a parent. Before doing that I think we can leave the warning there as a reminder.

PR #1333, opened 2018-01-18
Avoid showing multiple Geolocation permission dialog

mTwTm:
Per testing from https://permission.site, with a simple
location request we see two fires of onGeolocationPermissionsShowPrompt
Let's ignore the later ones if one is already shown.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -119,6 +119,7 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
     //GeoLocationPermission
     private String geolocationOrigin;
     private GeolocationPermissions.Callback geolocationCallback;
+    private AlertDialog geoDialog;
 
     /**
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -636,6 +637,14 @@ private void showGeolocationPermissionPrompt() {
         if (geolocationCallback == null) {
             return;
         }
+        if (geoDialog != null && geoDialog.isShowing()) {
+            return;
+        }
+        geoDialog = buildGeoPromptDialog();
```
idiotmax: What is the lifecycle of the geoDialog?
A Fragment holds the dialog, and the dialog holds the activity context. Is it possible that recreated activity reuse the fragment and hold the dialog to the activity cause a context leak?

mTwTm: Yeah, you're right. I pushed a new version while I'm not 100% sure this will be leak free, can you help take a look at it?

PR #1341, opened 2018-01-20
Make some static fields final

gaul:
Adding final prohibits modification and allows initialization of
primitive and String fields at compile time instead of runtime in
clinit:

https://developer.android.com/training/articles/perf-tips.html#UseFinal

Found via error-prone.


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotViewerActivity.java
```diff
@@ -69,11 +69,11 @@
     private static final int ACTION_SHARE = ACTION_VIEW + 2;
     private static final int ACTION_DELETE = ACTION_VIEW + 3;
 
-    private static int REQUEST_CODE_VIEW_SCREENSHOT = 101;
```
mTwTm: These REQUEST_CODE_VIEW_SCREENSHOT are actually not used any more in a recent patch, I should have removed that :P but yeah that's another issue.

PR #1347, opened 2018-01-23
[SurveyNotification] Use app icon as the large icon on survey notific…

pharspher:
…ation.


pharspher:
background color for small icon: #002275
screen shot 2018-01-23 at 6 22 14 pm

PR #1350, opened 2018-01-24
remove DS_Store for git on Mac OS X

Conjuror:
just to remove .DS_Store to avoid .DS_Store appear on git status


Conjuror:
This should be a better solution for it.

https://stackoverflow.com/questions/18393498/gitignore-all-the-ds-store-files-in-every-folder-and-subfolder/19299889#19299889

PR #1354, opened 2018-01-25
Fix window leakage occurs after pressing back key and menu button con…

pharspher:
…secutively.


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -835,6 +839,9 @@ public void onBackPressed() {
         if (this.mediator.handleBackKey()) {
             return;
         }
+
+        dismissBottomMenu();
```
mTwTm: Will it be easier to do this in onDestroy() ?

PR #1362, opened 2018-01-26
Nevin/espresso

cnevinc:
Add a simple Espresso test for ./gradlew connectedAndroidTest


cnevinc:
The purpose of this PR is to test CI with Espresso. Please don't merge this for now.
Todo list for this patch
1. Add a useful test and corresponding test uitls ( robot ) for test items like   
    a. Load a wepage
    b. Click on a menu item.
2. Add the ability to test a static web page ( maybe with MockWebServer)
3. Basic setup and tearDown method/ helper for others to use.

Possible stuff to note
1. We are making changes to build flavor so above task 2) may be affected.
2. I deleted previous tests because most of our UI is different than Focus. We might need to consider if we want to add them back.
3. Uiautomator is better for black box testing. Though I think it's useful for some test cases ( like sharing or giving feedback). I've left it in build.gradle but not using them cause I haven't get a use case for it. ( may need to remove it from PR if we won't use it here)

cnevinc:
After some discussion, we'll land this first to avoid CI check fail. I've created https://github.com/mozilla-tw/Rocket/issues/1367 for follow up

PR #1366, opened 2018-01-29
Try catch loadWebView version

idiotmax:


app/src/main/java/org/mozilla/focus/utils/DebugUtils.java
```diff
@@ -17,12 +17,23 @@ private DebugUtils() {
     }
 
     public static String loadWebViewVersion(Context context) {
-        return loadWebViewVersion(new WebView(context));
+        String webViewVersion;
+        try {
+            webViewVersion = loadWebViewVersion(new WebView(context));
+        } catch (Exception exception) {
+            webViewVersion = "";
```
mTwTm: Let's call this NO_WEBVIEW_VERSION = ""?

app/src/main/java/org/mozilla/focus/utils/DebugUtils.java
```diff
@@ -17,12 +17,23 @@ private DebugUtils() {
     }
 
     public static String loadWebViewVersion(Context context) {
-        return loadWebViewVersion(new WebView(context));
+        String webViewVersion;
```
mTwTm: Will
```
        try {
            return loadWebViewVersion(new WebView(context));   
        } catch (Exception ex) {
            return "";
        }
```
be simpler?

idiotmax: It's kinda a habit for debugging purpose. Do you agree?

PR #1368, opened 2018-01-29
add pre scripts and post scripts

Conjuror:
https://developer.nevercode.io/docs/moving-from-buddybuild-to-nevercode

### DIFF
--- a/NC_POST_BUILD_SCRIPT.sh
+++ b/NC_POST_BUILD_SCRIPT.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+set -e # Exit (and fail) immediately if any command in this scriptfails
+
+# Check that our APK files are not bigger than they should be
+python tools/metrics/apk_size.py
+
+# skip CODECOV
+#bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
--- a/NC_PRE_BUILD_SCRIPT.sh
+++ b/NC_PRE_BUILD_SCRIPT.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -e # Exit (and fail) immediately if any command in this scriptfails
+
+python tools/l10n/check_translations.py


NC_POST_BUILD_SCRIPT.sh
```diff
@@ -0,0 +1,24 @@
...
+# their SDK, and partly to allow selecting flavours in the BuddyBuild UI).
+# We don't know where the Buddybuild SDK lives, which causes gradle builds
+# to be broken (due to the SDK dependency injected into FocusApplication),
+# it's easiest just to revert to a clean source state here:
+git reset --hard
+
+# buddybuild doesn't seem to offer any direct way of running findbugs.
+# findbugs is run as part of |gradle build| and |gradle| check, but those
+# aren't run directly in buddybuild.
+./gradlew findbugs
```
mTwTm: `gradle :app:build` already covers findbugs and checkstyle, we didn't have the chance to do this back in BuddyBuild (Only `gradle :app:assemble` was available) but now we can choose which task to run on Nevercode so proabably we can skip these.

We have to deal with the Adjust key, though, if we want to go with the `gradle :app:build` plan.

NC_POST_BUILD_SCRIPT.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Check that our APK files are not bigger than they should be
+python tools/metrics/apk_size.py
+
+# buddybuild modifies our buildscripts and sources (this is partly to enable
+# their SDK, and partly to allow selecting flavours in the BuddyBuild UI).
+# We don't know where the Buddybuild SDK lives, which causes gradle builds
+# to be broken (due to the SDK dependency injected into FocusApplication),
+# it's easiest just to revert to a clean source state here:
+git reset --hard
```
mTwTm: We don't need that reset any more since we have no Buddybuild SDK now.
Also maybe we'd like to remove the comments and ./gradlew jacocoTestReport.

PR #1385, opened 2018-02-06
Implementation of data access object (using Room) to access multi-tabs persistent model

benjamin-cheng:


mTwTm:
Buddybuild is not building this pr, I'll try close and re-open.

PR #1387, opened 2018-02-07
Implementation of data access object (using Room) to access multi-tabs persistent model

benjamin-cheng:


mTwTm:
The buddybuild failure (Robolectric does not support API level 26.) is fixed in #1303 so this might depend on that.

PR #1391, opened 2018-02-07
Try to avoid potential memory leakage in webview

benjamin-cheng:
Reference: https://stackoverflow.com/questions/45744881/activity-has-leaked-intentreceiver-lollipopbrowseraccessibilitymanager/48596543#48596543


mTwTm:
The build seems to fail because Nevercode is not applying environment variables (Probably because of this being an external pull request.)
I'm trying to get this fixed :P

PR #1396, opened 2018-02-08
Bump 3000 instead of 2000 on nevercode

mTwTm:
Houston we have too many builds


Conjuror:
Per offline discuss, we should apply a better versioning counter. I'll give a new patch for the bumper later.

PR #1397, opened 2018-02-08
use version name to control versioning.

Conjuror:
This is a proposal for replacing #1396

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -40,10 +40,12 @@ android {
 
         def nevercode_build = System.getenv()["NEVERCODE_BUILD"]
         if (nevercode_build) {
-            // bump 2000 to avoid build number collision with buddybuild
-            def nevercode_build_number = String.valueOf(System.getenv()["NEVERCODE_BUILD_NUMBER"].toInteger() + 2000)
-            versionCode nevercode_build_number.toInteger()
-            versionNameSuffix "(" + nevercode_build_number + ")"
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -40,10 +40,12 @@ android {
 
         def nevercode_build = System.getenv()["NEVERCODE_BUILD"]
         if (nevercode_build) {
-            // bump 2000 to avoid build number collision with buddybuild
-            def nevercode_build_number = String.valueOf(System.getenv()["NEVERCODE_BUILD_NUMBER"].toInteger() + 2000)
-            versionCode nevercode_build_number.toInteger()
-            versionNameSuffix "(" + nevercode_build_number + ")"
+            if (System.getenv()["NEVERCODE_BRANCH"] == "master") {
+                def build_number = versionName.replaceAll("\\.", "").toInteger();
```
mTwTm: Note that with this change we will suddenly build "older" versions when we switch from 1.0.10 -> 1.1.0
I'm thinking if it would be better if we can fail the build when the digit count is > 3 so we will remember to change this.

Conjuror: I have two proposals here.

1. Add a condition check to make sure we have all three versions/sub-versions with one digit only.
2. Add a comment at the versionName above to remind the person who want to upgrade the version name.

I prefer the second one as an easy solution.

mTwTm: Oh yeah sure I was thinking of writing comment at this block was not enough but how can I not imagine writing comments before version...

Sounds like a plan to me!

PR #1398, opened 2018-02-08
Nevercode versioning

Conjuror:
It seems nevercode has problems with my force push. ORZ
This is the same as the last pull request of #1397


mTwTm:
The build is failing because Nevercode doesn't provide environment variable for forked pull requests. I'm trying to fix this and will re-spin a build when this is fixed.

mTwTm:
Not sure if we're still merging this. Closing for now. Please re-open if necessary.

PR #1399, opened 2018-02-08
Menu 2.0

mTwTm:


mTwTm:
Summary: Do a very quick implementation of 2.0 menu while Open-Tab-Tray is still home and function as home.

FIXME: The telemetry of v2 capture is sharing the old telemetry ping. Not sure if this should be fixed.

mTwTm:
Oops I did a force push to v2.0-dev, I'm going to lock it now. The range of this pr is 47d002a ... 08e7398

PR #1408, opened 2018-02-09
Try to avoid potential memory leakage in webview

benjamin-cheng:


mTwTm:
Target should be reset to v1.0-dev and maybe we want this in master (2.0) too?

PR #1415, opened 2018-02-09
Add Socket tag for search suggestion request

### LINKED ISSUE #1413: Crash when trying to search anything. (On Nightly)
Root cause: stricter regulation on network requests after upgrading sdk version.
This is definitely a S1 P1 I suppose.

mTwTm:
Fixes #1413

### DIFF
--- a/app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
+++ b/app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
@@ -5,6 +5,7 @@
 
 package org.mozilla.focus.urlinput;
 
+import android.net.TrafficStats;
 import android.os.AsyncTask;
 import android.support.annotation.MainThread;
 import android.support.annotation.NonNull;
@@ -12,6 +13,7 @@ import android.text.TextUtils;
 
 import org.json.JSONArray;
 import org.json.JSONException;
+import org.mozilla.focus.network.SocketTags;
 import org.mozilla.focus.search.SearchEngine;
 import org.mozilla.focus.utils.UrlUtils;
 
@@ -20,6 +22,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.ref.WeakReference;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
```diff
@@ -71,25 +74,36 @@ public void onInput(@NonNull CharSequence input, boolean isThrottled) {
...
-                }
-            }.execute(new URL(searchEngine.buildSearchSuggestionUrl(input.toString())));
-        } catch (MalformedURLException ex) {
-            // Do nothing
+
+    }
+
+    private static class QueryTask extends AsyncTask<String, Void, List<CharSequence>> {
+
+        private SoftReference<UrlInputContract.View> viewSoftReference;
```
benjamin-cheng: Do you intent to use `WeakReference`?

app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
```diff
@@ -71,25 +74,36 @@ public void onInput(@NonNull CharSequence input, boolean isThrottled) {
...
+
+        QueryTask(UrlInputContract.View view) {
+            viewSoftReference = new SoftReference<>(view);
         }
 
+        @Override
+        protected List<CharSequence> doInBackground(String... strings) {
+            TrafficStats.setThreadStatsTag(SocketTags.SEARCH_SUGGESTION);
```
benjamin-cheng: Should we tag the traffic only when the strict mode is enabled?

app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
mTwTm: Yes, to avoid memory leak.

app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
mTwTm: Is there any bad effect of tagging the traffic here? From doc I've read that tagging is a good practice for analyzing your traffics

app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
mTwTm: Oh ooops I see what you mean

mTwTm: I'll fix it later today.

app/src/main/java/org/mozilla/focus/urlinput/UrlInputPresenter.java
benjamin-cheng: Just wonder if the information is needed in release build. However, it's ok for me to keep tag of traffic statistic since it should be a lightweight operation.

PR #1417, opened 2018-02-09
Change apk size limit to 2.5MB

mTwTm:


Conjuror:
if this is the decision, why don't we simply merge it?

mTwTm:
@Conjuror I wasn't sure if we want this to be a blocker (on Nightly or on release). However I think we don't have this script run in Bitrise yet?

mTwTm:
We still need to trigger this on bitrise

Issue #100, opened 2017-08-01
[Compatibility] Cannot load images from Alodokter.com


Conjuror:
All images from [Alodokter.com](Alodokter.com) cannot be loaded by zerda.

[Alodokter.com](Alodokter.com) is no. 30 of Top 100 ID websites from Alexa

mTwTm:
Images seems to fall into the category of ad trackers. (Pure webview or disabling ad trackers in Settings may fix the problem)

harly21:
This is more related to WebCompat issue, remove UX&VD tag

Conjuror:
On Slack:

> tyu: This issue had UX&VD because we want to discuss about how might we hint user that she should disable speed mode to make things work.

Conjuror:
related to #342

Conjuror:
Close all web copmat issues to #342.

PR #1420, opened 2018-02-12
Try to resolve the intermittent UI test failures

benjamin-cheng:
`org.mozilla.focus.activity.MainActivityTest > testAppStart[nc-emulator(AVD) - 8.1.0] FAILED 
java.lang.RuntimeException: Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus. Root:
Root{application-window-token=android.view.ViewRootImpl$W@b3b9472, window-token=android.view.ViewRootImpl$W@b3b9472, has-window-focus=false, layout-params-type=1, layout-params-string=WM.LayoutParams{(0,0)(fillxfill) sim=#100 ty=1 fl=#81810100 pfl=0x20000 wanim=0x10302f6 needsMenuKey=2 colorMode=0}, decor-view-string=DecorView{id=-1
, visibility=VISIBLE, width=1080, height=1920, has-focus=true, has-focusable=true, has-window-focus=false, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=WM.LayoutParams{(0,0)(fillxfill) sim=#100 ty=1 fl=#81810100 pfl=0x20000 wanim=0x10302f6 need
sMenuKey=2 colorMode=0}, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0, child-count=2}}
:app:connectedFocusWebkitDebugAndroidTest FAILED`

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -548,3 +548,12 @@ task checkstyle(type: Checkstyle) {
 afterEvaluate {
     check.dependsOn 'findbugs', 'pmd', 'checkstyle'
 }
+
+task captureDeviceScreen(type: Exec) {
+    workingDir '..'
+    commandLine 'sh', 'tools/screenCapture.sh'
+}
+
+afterEvaluate {
+    connectedAndroidTest.dependsOn captureDeviceScreen
+}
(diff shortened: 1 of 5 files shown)


benjamin-cheng:
We move the CI to Bitrise and don't need to debug Nevercode's UI testing issue. Close this PR.

PR #1436, opened 2018-02-21
Nevin/firebase 105

cnevinc:
For collaboration:
Although this approach is different with Firebase's document [1], it works better for us cause we don't want developers to copy server side configurations to their sandbox projects. 

External contributors can play with their sandbox projects and update google-service.json accordingly. But there's no guarantee that they'll work as expected ( should fall back to default behavior if implemented correctly.)

[1] [Large team, safer development ](https://firebase.googleblog.com/2016/08/organizing-your-firebase-enabled-android-app-builds.html)


cnevinc:
btw, I left google-service.json in the repository cause this post: https://groups.google.com/forum/#!topic/firebase-talk/bamCgTDajkw

cnevinc:
The build in CI should fail. Cause I'll need to add package name(org.mozilla.rocket) and hash to Firebase console. I also need to update google-services.json to target the correct project as well.

cnevinc:
TODO: Separate google-service.json into different variant.
Currently, I only setup nevercode's debug package name (org.mozilla.rocket.debug.greenhouse ) for firebase. So others will fall.

PR #1451, opened 2018-02-28
Nevin/firebase 105 0301

cnevinc:
For collaboration:
Although this approach is different with Firebase's document [1], it works better for us cause we don't want developers to copy server side configurations to their sandbox projects. 

External contributors can play with their sandbox projects and update google-service.json accordingly. But there's no guarantee that they'll work as expected ( should fall back to default behavior if implemented correctly.)

[1] [Large team, safer development ](https://firebase.googleblog.com/2016/08/organizing-your-firebase-enabled-android-app-builds.html)


cnevinc:
I only enable Bitrise with Firebase project. For other CIs they'll fail ( cause they have different package names ( e.g. org.mozilla.rocket.debug.xxxx, which needs to be added manually to Firebase project.)

PR #1464, opened 2018-03-02
Restore previous open tabs on app launch

benjamin-cheng:


benjamin-cheng:
Use https://github.com/mozilla-tw/Rocket/pull/1468

PR #1465, opened 2018-03-02
bug: fix wrong checking statement

walkingice:
it caused by my bad git-rebase


walkingice:
The test fail is not related to this PR, ignore it for now

PR #1467, opened 2018-03-02
Modify MainMediator#showBrowserScreen() signature

pharspher:
The 2nd param of MainMediator#showBrowserScreen() is changed from "clearHistory" to "openInNewTab".

Why clearHistory param:
In prev version, history is always cleared if ACTION_VIEW is sent from 3rd party apps. However, we can't do this if ACTION_VIEW is sent by our self (e.g. open our feedback page, etc), which will make user unable to navigate back to the previous page.

With multi-tab supported, we can decide to open url in a new tab without clearing browsing stack. So clearHistory is not needed anymore.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -219,6 +219,7 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
                 }
             }
 
+
             private void actionDownloadGranted(Parcelable parcelable) {
                 Download download = (Download) parcelable;
                 queueDownload(download);
@@ -447,7 +448,7 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
         // restore WebView state
         if (savedInstanceState == null) {
             if (!TextUtils.isEmpty(pendingUrl)) {
-                loadUrl(pendingUrl);
+                loadUrl(pendingUrl, true);
                 pendingUrl = null;
             }
         } else {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
```diff
@@ -33,6 +33,7 @@
     private static final String MARKET_INTENT_URI_PACKAGE_PREFIX = "market://details?id=";
     private static final String EXTRA_BROWSER_FALLBACK_URL = "browser_fallback_url";
     public static final String EXTRA_IS_INTERNAL_REQUEST = "is_internal_request";
```
mTwTm: We always add `EXTRA_IS_INTERNAL_REQUEST` flag whenever we add `EXTRA_OPEN_NEW_TAB` flag. Can we get rid of `EXTRA_IS_INTERNAL_REQUEST` and make some change in 
```
private boolean isStartedFromExternalApp() {
```
with a new function 
```
private boolean isFromInternal() {
    return intent != null && intent.getBooleanExtra(IntentUtils.EXTRA_IS_INTERNAL_REQUEST, false); 
}
```
and write some comment there to illustrate this?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -50,6 +50,7 @@
 import android.webkit.WebBackForwardList;
 import android.webkit.WebChromeClient;
 import android.webkit.WebHistoryItem;
+import android.webkit.WebIconDatabase;
```
walkingice: ???

pharspher: I've removed unneeded import statement and force pushed

PR #1468, opened 2018-03-05
Restore previous open tabs on app launch

benjamin-cheng:


app/src/main/java/org/mozilla/focus/activity/MainMediator.java
```diff
@@ -151,6 +158,20 @@ private FragmentTransaction prepareBrowsing(@NonNull String url, boolean clearHi
...
+        final FragmentManager fragmentMgr = this.activity.getSupportFragmentManager();
+        FragmentTransaction transaction = fragmentMgr.beginTransaction();
+
+        final BrowserFragment browserFrg = (BrowserFragment) fragmentMgr
+                .findFragmentByTag(BrowserFragment.FRAGMENT_TAG);
+
+        if (browserFrg == null) {
+            final Fragment freshFragment = this.activity.createBrowserFragmentForRestoreTab(tabId);
+            transaction.replace(R.id.container, freshFragment, BrowserFragment.FRAGMENT_TAG);
+        }
```
walkingice: is is possible browserFrg is not null?

app/src/main/java/org/mozilla/focus/activity/MainMediator.java
```diff
@@ -63,6 +63,14 @@ public void showUrlInput(@Nullable String url) {
...
 
     public void showBrowserScreen(@NonNull String url, boolean clearHistory) {
+        clearInputAndCommitBrowserFragment(prepareBrowsing(url, clearHistory));
+    }
+
+    public void showBrowserScreenForRestoreTabs(@NonNull String tabId) {
+        clearInputAndCommitBrowserFragment(prepareBrowsingForRestoreTabs(tabId));
+    }
+
+    private void clearInputAndCommitBrowserFragment(FragmentTransaction browserFragmentTransaction) {
```
walkingice: I feel like intention of this method is not clear. Actually its implementation is to 1) clear input fragment 2) commit a transaction.

Here comes a question: what will it happen if we `browserFragmentTransaction.commit()` ? We have no clue about it. I suggest to change this method to `clearInputFragmentImmediate()`, then use this method in `showBrowserScreen` and `showBrowserScreenForRestoreTabs` respectively.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -405,6 +422,8 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         final View menuBtn = view.findViewById(R.id.btn_menu);
         if (tabCounter != null) {
             tabCounter.setOnClickListener(this);
+            TabsSession tabsSession = TabsSessionProvider.getOrNull(getActivity());
+            tabCounter.setCount(tabsSession != null ? tabsSession.getTabsCount() : 0);
```
walkingice: I guess this should be placed in another commit? and there is `TabsSessionProvider.getOrThrow` in below, I think we should reuse it.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -710,6 +716,11 @@ public BrowserFragment createBrowserFragment(@NonNull String url) {
         return fragment;
     }
 
+    public BrowserFragment createBrowserFragmentForRestoreTab(@NonNull String tabId) {
```
walkingice: In my personal opinion, this is not only for restoring tabs. It is actually for **creating a browser fragment and switch to specific tab**.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
benjamin-cheng: OK, I will move it to another commit. Thank you!

PR #1479, opened 2018-03-05
Restore to previous focused tab

benjamin-cheng:


app/src/main/java/org/mozilla/focus/tabs/TabModelStore.java
```diff
@@ -40,27 +42,34 @@ public static TabModelStore getInstance(@NonNull final Context context) {
...
                 listener.onSaveComplete();
             }
             return;
         }
 
+        PreferenceManager.getDefaultSharedPreferences(context)
+                .edit()
+                .putString(context.getResources().getString(R.string.pref_key_current_tab_id), currentTabId)
+                .apply();
+
```
walkingice: I am not sure that whether Preference is a good place to persist small piece of data, or to persist in DB. It is up to you.

app/src/main/java/org/mozilla/focus/tabs/TabsSession.java
```diff
@@ -85,15 +85,16 @@ public int getTabsCount() {
...
         for (final TabModel model : models) {
             final Tab tab = new Tab(model);
             bindCallback(tab);
             tabs.add(tab);
         }
 
         if (tabs.size() > 0 && tabs.size() == models.size()) {
-            currentIdx = 0; // first tab
+            int currentTabIndex = getTabIndex(currentTabId);
+            currentIdx = currentTabIndex != -1 ? currentTabIndex : 0;
```
walkingice: `(currentTabIndex != -1)` would be better. Since it is just a local variable in a small scope, we can make it shorter to be `final int index = getTabIndex(currentTabId);`

benjamin-cheng:
It has been refined according to the comments. Thank you! @walkingice

PR #1481, opened 2018-03-05
Add Mainmediator state awarness to LOAD_OPEN_URL

### LINKED ISSUE #1480: HOME -> URL BAR -> new page is not opened in a new tab
STR:
1. Tap Home
2. Tap url bar
3. Enter address and tap enter

Expected behavior:
New page opened in new tab

Actual behavior:
New page replaced old page

mTwTm:
Fixes #1480

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainMediator.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainMediator.java
@@ -62,6 +62,11 @@ public class MainMediator {
         this.prepareUrlInput(url).addToBackStack(UrlInputFragment.FRAGMENT_TAG).commit();
     }
 
+    // If openInNewTab is not provided, we decide based on current state of MainMediator
+    public void showBrowserScreen(final @NonNull String url) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/activity/MainMediator.java
```diff
@@ -62,6 +62,11 @@ public void showUrlInput(@Nullable String url) {
         this.prepareUrlInput(url).addToBackStack(UrlInputFragment.FRAGMENT_TAG).commit();
     }
 
+    // If openInNewTab is not provided, we decide based on current state of MainMediator
+    public void showBrowserScreen(final @NonNull String url) {
```
mTwTm: @walkingice I and @pharspher were discussing if it is appropriate for the name showBrowserScreen to handle based on isHomeFragmentVisible(). (Can we rename this to make it more explicit that this operation/API is based on state of isHomeFragmentVisible or MainMediator)

PR #1482, opened 2018-03-05
query serial sample

Conjuror:
This is a prototype of version query tool and use Google Spreadsheet as database.


Conjuror:
We'll need to add a few codes on pre-build script stage
`curl $BITRISEIO_SERVER_SECRET_JSON_URL`
`python tools/build_scripts/query_serial.py -c`

There's no solid way to pass build version. A quick solution is to use os.environ to pass the build number.

Conjuror:
Following libraries is required
`pip install --upgrade google-api-python-client google-auth oauth2client`

Conjuror:
be reminded to clean local build.serial before query

Conjuror:
Max and I were discussed about if we should get versionName from somewhere other than input manually. Therefore, we'll save the resource on maintaining build process. Here's an ugly proposal but definitely work.
```
            # with open("app/build.gradle") as f:
            #     for line in f:
            #         if 'versionName' in line:
            #             match = re.search('\d\.\d\.\d', line)
            #             if match:
            #                 args.v = match.group(0)
```

Conjuror:
Update: The status seems good to us to go. So here's what's happening next:

1. We'll use primary workflow (Pull Request workflow) to verify the proposal. The first is to dump versioning call to the console
2. If step 1 goes correct, we'll send out an updated pull request to replace current versioning
3. Then, we'll change back primary and use the proposal for deploy_nightly workflow and release workflow

Conjuror:
So, after the last update, all the confirmation is done and we are almost ready for landing it. The last thing we have to do is to update the Rocket Build Map sheet to the correct version code. Currently, the version code is generated manually. Please make sure the version code is good before land this patch.

mTwTm:
No longer relevant.

PR #1517, opened 2018-03-07
[TabTray] Go to home fragment after the last tab is closed.

pharspher:


app/src/main/java/org/mozilla/focus/tabs/tabtray/TabTrayFragment.java
```diff
@@ -391,6 +392,20 @@ private void updateWindowAttrs(@NonNull Window window) {
...
     }
 
+    private void notifyFragmentListener(@NonNull FragmentListener.TYPE type, @Nullable Object payload) {
+        Activity activity = getActivity();
+        if (activity instanceof FragmentListener) {
+            ((FragmentListener) activity).onNotified(this, type, payload);
+        }
+    }
+
+    private Runnable dismissRunnable = new Runnable() {
```
walkingice: can we move this declaration to upper region of this file? also, there is similar code in `tabSwitched`, and you might like to refactor it as well.

pharspher: I've moved the declaration of dismiss runnable to the top of the class, also I remove the duplicate code in tabSwitched(). Thank you.

PR #1523, opened 2018-03-08
Issues/1522

### LINKED ISSUE #1522: For non-external intialized App launches, we should show Home before we leave.

mTwTm:
Fixes #1522
BrowserFragment is now sharing the lifecycle to MainActivity.
In the meanwhile the TabsSession destroy is delayed to:
1. Match the creation lifecycle (MainActivity)
2. Avoid too aggressive destroy which make users might need to reload all pages (when BrowserFragment is poped.)


mTwTm:
Need some more work on this branch. Tabtray is not working if BrowserFragment is popped.

mTwTm:
UI test fail reproduce locally on API 21 emulator (but not on 27, strange.) The test hangs and do nothing and crash at the end. Have not found the root cause yet and failed to work-around. The test fail is beginning from 60128eb

mTwTm:
Deprecated, favoring #1590

Issue #1424, opened 2018-02-12
Run experiements to promote "Share with friends"


joechengla:
and understand its impact on retention / acquisition by referral

Jacklin-ux:
The spec link:
https://drive.google.com/file/d/1kI1pZfYU6MEnAlFAzXCq5ZHQjhgz04DM/view?usp=sharing

Issue #1423, opened 2018-02-12
Run experiments to drive users to "set default browser"


joechengla:
and understand its impact on retention

Jacklin-ux:
The spec link:
https://drive.google.com/file/d/1kI1pZfYU6MEnAlFAzXCq5ZHQjhgz04DM/view?usp=sharing

PR #1524, opened 2018-03-08
Nevin/promotion

cnevinc:
Run experiments to promote "Share with friends" #1424
Run experiments to promote "Default browser" #1423


cnevinc:
resolve #1424  #1423

PR #1536, opened 2018-03-12
[TabTray] Refine UI of tab item

pharspher:


app/src/main/java/org/mozilla/focus/tabs/tabtray/TabTrayAdapter.java
```diff
@@ -39,8 +41,13 @@ public ViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
     public void onBindViewHolder(ViewHolder holder, int position) {
         holder.itemView.setSelected(position == this.focusedTabPosition);
 
+        Resources resources = holder.itemView.getResources();
         Tab tab = tabs.get(position);
-        holder.websiteTitle.setText(tab.getTitle());
+
+        String title = tab.getTitle();
+        holder.websiteTitle.setText(TextUtils.isEmpty(title) ?
+                resources.getString(R.string.app_name) : title);
```
benjamin-cheng: Suppose it's the decision from design team right? To show app name instead when the title is empty. However, I think it could be more reasonable to show "No title" instead of the app name.

PR #1553, opened 2018-03-13
[TabTray] Implement favicon

pharspher:


app/src/main/java/org/mozilla/focus/tabs/tabtray/TabTrayAdapter.java
```diff
@@ -76,18 +93,64 @@ int getFocusedTabPosition() {
         return this.focusedTabPosition;
     }
 
+    private void loadFavicon(Tab tab, final ViewHolder holder) {
+        RequestOptions options = new RequestOptions()
+                .diskCacheStrategy(DiskCacheStrategy.RESOURCE)
```
benjamin-cheng: Have you checked the default disk cache size?

PR #1556, opened 2018-03-13
tabs: add more information to onTabHoist

walkingice:
Add factor to indicate potential reason which cause the Tab-hoist event.


app/src/main/java/org/mozilla/focus/tabs/TabsChromeListener.java
```diff
@@ -22,6 +22,13 @@
  */
 public interface TabsChromeListener {
 
+    enum Factor {
```
pharspher: What about using IntDef instead of enum to avoid increasing to much size to the dex?

PR #1559, opened 2018-03-13
Save/restore web view state of the tabs part 1

benjamin-cheng:


app/src/main/java/org/mozilla/focus/tabs/TabModelStore.java
```diff
@@ -67,15 +71,25 @@ public void saveTabs(@NonNull final Context context,
...
+                Context context = contextRef.get();
+                if (context != null) {
+                    File cacheDir = new File(context.getCacheDir(), TAB_WEB_VIEW_STATE_FOLDER_NAME);
+                    for (TabModel tabModel : tabModelList) {
+                        if (tabModel != null) {
+                            tabModel.setWebViewState(FileUtils.readBundleFromStorage(cacheDir, tabModel.getId()));
+                        }
+                    }
+                }
+                return tabModelList;
```
walkingice: Could we refactor this part since it is a bit deep.

app/src/main/java/org/mozilla/focus/tabs/TabModelStore.java
```diff
@@ -109,6 +125,16 @@ protected Void doInBackground(TabModel... tabModelList) {
...
+                    Context context = contextRef.get();
+                    if (context != null) {
+                        File cacheDir = new File(context.getCacheDir(), TAB_WEB_VIEW_STATE_FOLDER_NAME);
+                        for (TabModel tabModel : tabModelList) {
+                            if (tabModel != null && tabModel.getWebViewState() != null) {
+                                FileUtils.writeBundleToStorage(cacheDir, tabModel.getId(), tabModel.getWebViewState());
+                            }
+                        }
+                    }
+
```
walkingice: refactor this?

app/src/main/java/org/mozilla/focus/utils/FileUtils.java
```diff
@@ -185,4 +186,18 @@ public static long clearCache(Context context) {
...
+        ensureDir(dir);
+        // TODO: Not implement yet
+    }
+
+    public static Bundle readBundleFromStorage(@NonNull final File dir,
+                                               @NonNull final String fileName) {
+        ensureDir(dir);
+        // TODO: Not implement yet
+        return null;
+    }
```
walkingice: In my personal view, 'ToStorage' and 'FromStorage' might be redundant, but keep it still OK to me.

Is this method designed to write/read general Bundle or specific to WebViewState only?

benjamin-cheng:
@walkingice  Thanks for your comments. I prefer to merge this PR and refine the code in the following PR.

app/src/main/java/org/mozilla/focus/utils/FileUtils.java
benjamin-cheng: They are general methods for persisting Bundle in the long run.

PR #1561, opened 2018-03-14
[Tab] Add Tab#getSecurityState() to check whether the connection is secure

pharspher:


app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -189,6 +190,11 @@ public void reload() {
         }
     }
 
+    @Override
+    public @SiteIdentity.SecurityState int getSecurityState() {
+        return getCertificate() == null ? SiteIdentity.INSECURE : SiteIdentity.SECURE;
+    }
+
```
walkingice: I am not pretty sure whether this logic is enough. Maybe we can add an issue to track this.

PR #1564, opened 2018-03-14
[TabTray] Do not update ui from background tab

pharspher:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -1288,9 +1296,34 @@ public void onReceivedTitle(@NonNull Tab tab, String title) {
...
+
+        private void onForegroundTabHoist(Tab tab) {
+            geolocationOrigin = "";
+            geolocationCallback = null;
+
+            dismissGeoDialog();
+
+            updateURL(tab.getUrl());
+
+            int identity = tab.getSecurityState() == SiteIdentity.SECURE ? SITE_LOCK : SITE_GLOBE;
```
walkingice: please `(tab.getSecurityState() ==  SiteIdentity.SECURE) ? SITE_LOCK : SITE_GLOBE;`

pharspher: updated and force-pushed

PR #1566, opened 2018-03-15
Serialize the webview state bundle to file

benjamin-cheng:


app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
```diff
@@ -0,0 +1,151 @@
...
+    public interface DataTypeHandler {
+        String getName();
+
+        SerializedItem create(Bundle bundle, String key) throws IOException;
+
+        void restore(Bundle bundle, SerializedItem serializedItem) throws IOException;
+    }
+
+    public AndroidBundleSerializer() {
+        dataTypeHandlers = new HashMap<>();
```
walkingice: we can initialize it in declaration
```
private final Map<String, DataTypeHandler> dataTypeHandler = new HashMap<>();
```

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
```diff
@@ -0,0 +1,151 @@
...
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class AndroidBundleSerializer {
+
+    private final Map<String, DataTypeHandler> dataTypeHandlers;
```
walkingice: I am not pretty sure, can we just use 'Class' as key for this map?

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public void serializeBundle(ObjectOutputStream outputStream, Bundle bundle) throws IOException {
+        if (bundle != null && bundle.size() > 0) {
+            List<SerializedItem> serializedItemList = new ArrayList<>();
+            Set<String> keys = bundle.keySet();
+            for (String key : keys) {
+                Object value = bundle.get(key);
+                if (value == null) {
+                    continue;
+                }
```
walkingice: In my personal opinion, this is not what I expect if we want to make it general
```java
Bundle before = new Bundle();
before.putString("keyA", "a");
before.putString("keyB", null);
before.contains("keyA"); // return true
before.contains("keyB"); // return true

Bundle after = deserialize(serialize(before));
after.contains("keyA"); // return true
after.contains("keyB"); // return false
```

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class AndroidBundleSerializer {
+
+    private final Map<String, DataTypeHandler> dataTypeHandlers;
+
+    public interface DataTypeHandler {
+        String getName();
```
walkingice: I suggest to use `getSimpleName` since the implementation do use `getSimpleName`

app/src/main/java/org/mozilla/focus/utils/SerializedItem.java
```diff
@@ -1,6 +1,7 @@
 package org.mozilla.focus.utils;
 
 import java.io.Serializable;
+import java.nio.ByteBuffer;
```
walkingice: please squash this whole commit into the original implementation of SerializedItem

app/src/main/java/org/mozilla/focus/tabs/Tab.java
```diff
@@ -180,6 +180,10 @@ public void detach() {
     }
 
     /* package */ TabView createView(@NonNull final Activity activity) {
+        return createViewWithFallbackUrl(activity, null);
+    }
+
+    /* package */ TabView createViewWithFallbackUrl(@NonNull final Activity activity, final String url) {
```
walkingice: I would suggest that rename `createView` to `initializeView`. We don't need pass url as parameter since the url do come from this tab itself.

```java
 /* package */ TabView initializeView(@NonNull final Activity activity) {
        final String url = this.getUrl();
```

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
benjamin-cheng: I prefer not to leak the implementation detail for this method.

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
benjamin-cheng: Use String as key can reduce traversal effort to check the exact type of the `SerializedItem`. If we use `Class` object as key and serialize to file, it would need to use `instanceof` to see if it's a byte[] or String class.

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
benjamin-cheng: For this case, it seems a NULL handler is needed to serialize/de-serialize the null value.

app/src/main/java/org/mozilla/focus/tabs/Tab.java
benjamin-cheng: Need to discuss with you.

app/src/main/java/org/mozilla/focus/utils/AndroidBundleSerializer.java
benjamin-cheng: OK

PR #1567, opened 2018-03-15
Julian/switch and add

walkingice:


app/src/main/java/org/mozilla/focus/tabs/TabsSession.java
```diff
@@ -41,11 +42,7 @@
 
     private Notifier notifier;
 
-    /**
-     * Index to refer a tab which is 'focused' by this session. When this index be changed, session
-     * should hoist focused tab as well.
-     */
-    private int focusIdx = -1;
+    private WeakReference<Tab> focusRef = new WeakReference<>(null);
```
benjamin-cheng: Why is it necessary to use WeakReference on the focus tab? Is there any scenario that could cause memory leakage?

walkingice: Frankly, there is no strong reason to use it, just wanna give a try on WeakReference

pharspher: i didn’t see any statement to assign new focused tab in removeTab(). With strong ref to the focused tab, if the user calls removeTab without any succeeding action to assign new focused tab, the original focused one may be leaked in this case? I am reading code on my phone, it’s possible I didn’t read carefully enough, sorry if this is the case.

Issue #56, opened 2017-07-26
[Function] Menu->Add to homescreen


chianger:
P2

Jacklin-ux:
The spec link:
https://drive.google.com/file/d/1LHDa3sF9DU-HjvcsSoPRqPBga4-XTu1K/view?usp=sharing

chianger:
merged

PR #1568, opened 2018-03-15
add2home: Pin Shortcut #56

cnevinc:
Add Pin Shortcut feature per #56

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.utils;
+
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentSender;
+import android.content.pm.ShortcutInfo;
+import android.content.pm.ShortcutManager;
+import android.graphics.Bitmap;
+import android.os.Build;
+import android.support.v4.content.pm.ShortcutInfoCompat;
+import android.support.v4.content.pm.ShortcutManagerCompat;
+import android.support.v4.graphics.drawable.IconCompat;
+import android.text.TextUtils;
+import android.util.Log;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ShortcutUtils {
+
+    private static final String TAG = "ShortcutUtils";
+
+
+    // Request pinned shortcut for both API level
+    public static void requestPinShortcut(final Context context, final Intent shortcutIntent,
+                                          final String title, final String urlAsShortcutId, final Bitmap bitmap) {
+
+        if (context == null || shortcutIntent == null || title == null) {
+            Log.e(TAG, "requestPinShortcut fail cause some required arguments are null.");
+            return;
+        }
+
+        Bitmap icon = bitmap;
+        if (icon == null) {
+            // if favicon is not ready, we use the default initial icon with white color
+            icon = FavIconUtils.getInitialBitmap(context.getResources(), null, FavIconUtils.getRepresentativeCharacter(urlAsShortcutId));
+        }
+        // label must not be empty
+        String label = title;
+        if (TextUtils.isEmpty(title)) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
```diff
@@ -0,0 +1,77 @@
...
+            final ShortcutManager shortcutManager = context.getSystemService(ShortcutManager.class);
+            if (shortcutManager != null) {
+                final List<ShortcutInfo> list = new ArrayList<>();
+                list.add(shortcut.toShortcutInfo());
+                shortcutManager.updateShortcuts(list);
+            }
+        }
+
+
+        ShortcutManagerCompat.requestPinShortcut(context, shortcut, intentSender);
```
idiotmax: What if the Launcher app does not support Shortcut Pinning?(isRequestPinShortcutSupported() return false)

app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // Display home screen after add to home screen
+        final Intent showHome = new Intent(Intent.ACTION_MAIN);
+        showHome.addCategory(Intent.CATEGORY_HOME);
+        showHome.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        final PendingIntent pendingIntent = PendingIntent.getActivity(context, 0, showHome, PendingIntent.FLAG_UPDATE_CURRENT);
+        final IntentSender intentSender = pendingIntent.getIntentSender();
+
+        // Update the shortcut icon on launcher since previous one may not ready. API 25+ only
+        // TODO: find a way to update the shortcut icon for API 25-. Currently the only way is remove old shortcut and add again.
```
idiotmax: If we do not allow duplicate, will this install treated as update?

https://searchfox.org/mozilla-central/rev/8976abf9cab8eb4661665cc86bd355cd08238011/mobile/android/base/java/org/mozilla/gecko/util/ShortcutUtils.java#51

app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
cnevinc: Currently, nothing will happen. I assume we might need to show some text for users. Or just disable the icon before it shows up

app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
cnevinc: For O+, the icon will be duplicated but using the old icon. So this is still a must have.
For O-, I can't get the install intent (INSTALL_SHORTCUT) like Fennec because it's generated inside of support library.
https://github.com/aosp-mirror/platform_frameworks_support/blob/b6838fd2d2e834fdd38aab34511d385cb7108f63/compat/src/main/java/android/support/v4/content/pm/ShortcutManagerCompat.java#L98

This fix I can think of is copy the code and edit it. But I'd like to use the system behaviour if I didn't think it wrong.

app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
```diff
@@ -0,0 +1,77 @@
...
+public class ShortcutUtils {
+
+    private static final String TAG = "ShortcutUtils";
+
+
+    // Request pinned shortcut for both API level
+    public static void requestPinShortcut(final Context context, final Intent shortcutIntent,
+                                          final String title, final String urlAsShortcutId, final Bitmap bitmap) {
+
+        if (context == null || shortcutIntent == null || title == null) {
```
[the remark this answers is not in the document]
cnevinc: sure

app/src/main/java/org/mozilla/focus/utils/ShortcutUtils.java
idiotmax: How about let's use ShortcutManagerCompat.createShortcutResultIntent to retrieve the Intent and patch the duplicate extra?

cnevinc:
ShortcutManagerCompat.createShortcutResultIntent can give us the intent used to pin shortcut. We can set duplicates to false there. However,  this won't update the icon.
If we set duplicate to true, then the new icon will be the latest one. The old one is still the same.

If we just call ShortcutManagerCompat.requestPinShortcut for API 24 and below, the behaviour will be the same as setting duplicates to true.

PR #1576, opened 2018-03-16
promotion: telemetry for #1424 and #1423

cnevinc:
Add telemetry for rate(#1424) and default browser(#1423) notification


cnevinc:
same patch as #1558

PR #1594, opened 2018-03-19
settings: Fix bug for crashes while setting default browser

cnevinc:


app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
```diff
@@ -231,7 +231,7 @@ public static void goToPlayStore(Context context) {
     }
 
 
-    public static Intent genDefaultBrowserSettingIntent(Context context) {
+    public static Intent genDefaultBrowserSettingIntentForBroadcastReceiver(Context context) {
```
mTwTm: We probably can remove that public now, too.

PR #1599, opened 2018-03-19
settings: Fix bug for crashes while setting default browser for 1.1

cnevinc:
Same as #1594 but targets v.1.0-dev

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
@@ -228,7 +228,7 @@ public class IntentUtils {
     }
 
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
```diff
@@ -228,7 +228,7 @@ public static void goToPlayStore(Context context) {
     }
 
 
-    public static Intent genDefaultBrowserSettingIntent(Context context) {
+    public static Intent genDefaultBrowserSettingIntentForBroadcastReceiver(Context context) {
```
mTwTm: We probably can remove that public now, too.

PR #1603, opened 2018-03-19
[BrowserFragment] Show a snack bar to tell that a new tab is loaded i…

### LINKED ISSUE #1552: There should be a snack bar to inform user that the new tab is added
STR:
1. Launch Rocket
2. Tap any top website
3. Long press a link
4. Tap "Open link in new tab"

Expected result:
 - There should be a snack bar to inform users that the new tab is added

Actual result: 
 - It doesn't show a snack bar

Device: Redmi Note4
Rocket nightly: 1.0.4 (2557). nightly
Android version: 7.0
WebView version 64.0.3282.137

pharspher:
…n the background, fix #1552

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -103,6 +103,11 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
         BackKeyHandleable, ScreenshotObserver.OnScreenshotListener {
 
     public static final String FRAGMENT_TAG = "browser";
+
+    /** Custom data that is passed when calling {@link TabsSession#addTab(String, Bundle)} */
+    public static final String EXTRA_NEW_TAB_SRC = "extra_bkg_tab_src";
+    public static final int SRC_CONTEXT_MENU = 0;
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -103,6 +103,11 @@
         BackKeyHandleable, ScreenshotObserver.OnScreenshotListener {
 
     public static final String FRAGMENT_TAG = "browser";
+
+    /** Custom data that is passed when calling {@link TabsSession#addTab(String, Bundle)} */
+    public static final String EXTRA_NEW_TAB_SRC = "extra_bkg_tab_src";
+    public static final int SRC_CONTEXT_MENU = 0;
```
walkingice: I don't have strong opinion: It also make sense to me if these two lines were placed in ContextMenu

pharspher: The reason I put it here is because 
1. BrowserFragment is the last receiver of the event, and is also the only one that knows how to react to this field. 
2. The number of sources may increase in the future(e.g. we might want to change UI when  a tab is added by javascript). Putting all these src definitions at one place can make maintenance easier.

PR #1612, opened 2018-03-20
Implement take screenshot test.

josephyu1110:
- screenshot_004

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/TakeScreenshotTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/TakeScreenshotTest.java
@@ -1,0 +1,135 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.activity;
+
+import android.Manifest;
+import android.content.Intent;
+import android.os.SystemClock;
+import android.support.annotation.Keep;
+import android.support.test.espresso.Espresso;
+import android.support.test.espresso.IdlingRegistry;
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/TakeScreenshotTest.java
```diff
@@ -0,0 +1,129 @@
...
+        // Click screen capture button
+        onView(allOf(withId(R.id.btn_capture), isDisplayed())).perform(click());
+
+        // Register screenshot taken idling resource and wait capture complete
+        IdlingRegistry.getInstance().register(screenshotIdlingResource);
+
+        IdlingRegistry.getInstance().unregister(screenshotIdlingResource);
+
+        // Wait a while to get animation dimiss
+        SystemClock.sleep(TIMEOUT_WAIT_ANIMATION_FINISHED);
```
cnevinc: We need to disable animation in Android system setting (developer option) first.
This method is not ideal but there is some issue with our CI server. so I hesitate to approve it.
https://developer.android.com/training/testing/espresso/setup.html

app/src/androidTest/java/org/mozilla/focus/activity/TakeScreenshotTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+        // Open menu
+        onView(allOf(withId(R.id.btn_menu), isDisplayed())).perform(click());
+
+        // Click my shot
+        onView(allOf(withId(R.id.menu_screenshots), isDisplayed())).perform(click());
+
+        // Click the first item in my shots panel
+        onView(withId(R.id.screenshot_grid_recycler_view)).perform(
+                RecyclerViewActions.actionOnItemAtPosition(1, click()));
```
cnevinc: Why it's not 0?

josephyu1110: Since "index=0" in ScreenShotItemAdapter is always date label, the first  screenshot item will start from "index=1".

PR #1613, opened 2018-03-20
[Menu] Disable forward/reload/share/shortcut button if browser fragme…

### LINKED ISSUE #1589: Rocket crashes when trying to reload the home panel after closing all tabs
1. Close all tabs in tab view
2. Now at home panel with zero tabs
3. Tap menu and tap reload or X
4. Rocket crashes

pharspher:
…nt is not in the foreground, fix #1589

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -636,19 +636,16 @@ public class MainActivity extends LocaleAwareAppCompatActivity implements Fragme
         // Only refresh when disabling turbo mode
         if (this.getResources().getString(R.string.pref_key_turbo_mode).equals(key)) {
             final boolean turboEnabled = isTurboEnabled();
-            BrowserFragment browserFragment = getVisibleBrowserFragment();
+            BrowserFragment browserFragment = getBrowserFragment();
             if (browserFragment != null) {
-                browserFragment.setBlockingEnabled(turboEnabled);
-                // Reload if we're closing Turbo mode since we should be fixing something
-                if (!turboEnabled) {
-                    browserFragment.reload();
-                }
+                browserFragment.setContentBlockingEnabled(turboEnabled);
             }
             menu.findViewById(R.id.menu_turbomode).setSelected(turboEnabled);
         } else if (this.getResources().getString(R.string.pref_key_performance_block_images).equals(key)) {
             final boolean blockingImages = isBlockingImages();
-            if (getVisibleBrowserFragment() != null) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -378,11 +378,7 @@ private void setUpMenu() {
     @VisibleForTesting
     public BrowserFragment getVisibleBrowserFragment() {
```
mTwTm: Seems like some functions that depends on getVisibleBrowserFragment() will change behavior: for instance, when TopFragment is HomeFragment and TurboMode is switched, Site currently focused will not be reloaded. (However this sounds acceptable, lol.)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -575,6 +575,12 @@ public void onDestroyView() {
         super.onDestroyView();
     }
 
+    public void setImageBlockingEnabled(boolean enabled) {
+        for (Tab tab : tabsSession.getTabs()) {
+            tab.setImageBlockingEnabled(enabled);
+        }
+    }
+
```
walkingice: it would be sweet if we add `TODO: this implementation should not belong to Fragment` to this method, as well as `setContentBlockingEnabled`

app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -137,14 +134,37 @@ public void saveViewState(Bundle outState) {
...
 
     public boolean isBlockingEnabled() {
         return client.isBlockingEnabled();
     }
 
+    @Override
+    public void setImageBlockingEnabled(boolean enabled) {
+        WebSettings settings = getSettings();
+        if (enabled == settings.getBlockNetworkImage() &&
+                enabled == !settings.getLoadsImagesAutomatically()) {
```
walkingice: please move `&&` to the start of next line. Maybe to use `flag` is better than `enabled`? (since enabled indicate a happened-state of WebSettings)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
pharspher: comment added

app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
pharspher: already moved the operator to the head of the second line.

PR #1628, opened 2018-03-21
Ignore WebGeolocationPermissionTest on devices api > 23

mTwTm:


mTwTm:
@josephyu1110 That's a very useful tip! Thanks! I've updated it accordingly.

cnevinc:
+1

PR #1632, opened 2018-03-21
Disable contextual feature hint on Screenshot

### LINKED ISSUE #1547: Disable contextual feature hint on Screenshot
Now that the screenshot feature is surfaced as a main feature. I suggest to disable the contextual hint for now

benjamin-cheng:
Fix #1547


app/src/main/res/values/strings.xml
```diff
@@ -279,9 +279,6 @@
     <string name="screenshot_image_viewer_dialog_info_url1">URL: %1$s</string>
     <string name="screenshot_image_viewer_dialog_delete_msg">Delete this image?</string>
 
-    <string name="screenshot_on_boarding_text_msg">Tap the screenshot button in the menu to capture the whole page.\nFast, easy - come back to the page at anytime!</string>
```
mTwTm: Just to let you know for strings that we may use later we don't remove them so that they are not "removed from Pontoon and have to be re-translated". I think we probably don't need these anymore so it's good to be removed

PR #1649, opened 2018-03-22
Fix transition animation between fragments

pharspher:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -847,17 +850,36 @@ public boolean onBackPressed() {
...
+     * @param url target url
+     * @param openNewTab whether to load url in a new tab or not
+     * @param onViewReadyCallback callback to notify that web view is ready for showing.
+     */
+    public void loadUrl(@NonNull final String url, boolean openNewTab,
+                        final Runnable onViewReadyCallback) {
+        // In case we call TabsSession#addTab(), which is an async operation calls back in the next
+        // message loop. By posting this runnable we can call back in the same message loop with
+        // TabsContentListener#onFocusChanged(), which is when the view is ready and being attached.
+        final Runnable callback = new Runnable() {
```
mTwTm: I'm uncertain why we wrap a Runnable with another Runnable here? Is this the intentional delay you mentioned earlier? However this look like nothing is delayed to me.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -120,6 +121,8 @@
 
     private static final int BUNDLE_MAX_SIZE = 300 * 1000; // 300K
 
+    private Handler uiHandler = new Handler(Looper.getMainLooper());
```
mTwTm: Maybe we can just use ThreadUtils.postToMainThread() if this is a one-time thing?

pharspher: Ok, I didn't notice there's such a util class, I'll change to use that. Thx!

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
pharspher: Embarrassingly I don't know what I'm doing here, completely nonsense. I'll fix that.

PR #1656, opened 2018-03-23
Import from l10n repo cb1714b (1.x)

mTwTm:


mTwTm:
I tend to merge this later since Github releases are branch based and we're targetting v1.0-dev this time rather then branching another 1.1.0

PR #1724, opened 2018-03-29
Gradle dependencies by Kotlin DSL

benjamin-cheng:
Kotlin + buildsrc = Android Studio Autocomplete & Click Support

Reference: https://handstandsam.com/2018/02/11/kotlin-buildsrc-for-better-gradle-dependency-management/


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.5.1-all.zip
```
cnevinc: Is this a must?

benjamin-cheng: No. It's ok to have this feature with gradle 4.1. However, the plugin and kotlin version(1.1.3) is pretty old in gradle 4.1. From gradle 4.5, the built-in kotlin version is upgraded to 1.2.0.

benjamin-cheng:
Yes. apply this mechanism doesn't increase the APK size.

mTwTm:
Yeah with the new added script we are confident that this change will not bump APK size (diff is around 1KB)

gradle/wrapper/gradle-wrapper.properties
mTwTm: I think we should update this. Just maybe not before release so maybe merge this after we build beta rc.

PR #1729, opened 2018-03-29
Julian/change close tab behavior

walkingice:
Do not close root-tab


walkingice:
We could do the work of changing-focus in another PR if necessary

PR #1737, opened 2018-03-30
Update telemetry for 2.0 multi-tab

### LINKED ISSUE #1509: Rocket 2.0 data/telemetry
After spec lock down we should figure out what telemetry are needed and how these should be embed in product.  Data review again will be necessary.

idiotmax:
Closes #1509

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
@@ -60,6 +60,7 @@ public final class TelemetryWrapper {
         private static final String EDIT = "edit";
         private static final String PERMISSION = "permission";
         private static final String FULLSCREEN = "fullscreen";
+        private static final String ADD = "add";
 
         private static final String FOREGROUND = "foreground";
         private static final String BACKGROUND = "background";
@@ -80,6 +81,8 @@ public final class TelemetryWrapper {
         private static final String CAPTURE = "capture";
         private static final String SEARCH_SUGGESTION = "search_suggestion";
         private static final String SEARCH_BAR = "search_bar";
+        private static final String TAB = "tab";
+        private static final String TABTRAY = "tab_tray";
 
         private static final String SETTING = "setting";
         private static final String APP = "app";
@@ -96,6 +99,7 @@ public final class TelemetryWrapper {
 
     public static class Value {
         private static final String HOME = "home";
+        private static final String TOPSITE = "top_site";
         private static final String DOWNLOAD = "download";
         private static final String HISTORY = "history";
         private static final String TURBO = "turbo";
@@ -103,6 +107,7 @@ public final class TelemetryWrapper {
         private static final String CLEAR_CACHE = "clear_cache";
         private static final String SETTINGS = "settings";
 
+        private static final String TABTRAY = "tab_tray";
         private static final String TOOLBAR = "toolbar";
         private static final String FORWARD = "forward";
         private static final String RELOAD = "reload";
@@ -141,6 +146,7 @@ public final class TelemetryWrapper {
         private static final String SUCCESS = "success";
         private static final String SNACKBAR = "snackbar";
         private static final String SOURCE = "source";
+        private static final String VERSION = "version";
     }
 
     public static class Extra_Value {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -332,6 +337,12 @@ public static void copyImageEvent() {
         TelemetryEvent.create(Category.ACTION, Method.COPY, Object.BROWSER_CONTEXTMENU, Value.IMAGE).queue();
     }
 
+    public static void addNewTabFromContextMenu() {
+        TelemetryEvent.create(Category.ACTION, Method.ADD, Object.TAB, Value.LINK)
```
mTwTm: I'm uncertain here: to be comparable with shareLinkEvent() it should be
TelemetryEvent.create(Category.ACTION, Method.OPEN_TAB, Object.BROWSER_CONTEXTMENU, Value.LINK).queue();
but to be comparable with addNewTabFromHome() it should be
TelemetryEvent.create(Category.ACTION, Method.ADD, Object.TAB, Value.LINK).queue();
I think I'm favoring the first one?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -453,12 +464,15 @@ public static void clickAddToHome() {
...
-        TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.TOOLBAR, Value.CAPTURE).queue();
+        TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.TOOLBAR, Value.CAPTURE)
+                .extra(Extra.VERSION, Integer.toString(2))
+                .queue();
     }
 
     public static void clickTopSiteOn(int index) {
         TelemetryEvent.create(Category.ACTION, Method.OPEN, Object.HOME, Value.LINK)
                 .extra(Extra.ON, Integer.toString(index))
+                .extra(Extra.VERSION, Integer.toString(2))
```
mTwTm: I'm actually uncertain why we upgrade this. (We have not changed it somehow?) Maybe we want to know if the new home / browser navigation has an impact but maybe we should upgrade more? Like Home / BrowserFragment first level item clicks?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
idiotmax: For easier quantitative comparison, do we focus on how user add a tab(add,tab,xxx)? Or which one is more popular when user long click on ContextMenu([share/copy/save],browser_contextmenu,[image/link])?

mTwTm: That sounds like a product call but I think of UI pings more related to UI itself so comparing these are a different dash board which we select different pings from different UIs.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -332,6 +338,12 @@ public static void copyImageEvent() {
         TelemetryEvent.create(Category.ACTION, Method.COPY, Object.BROWSER_CONTEXTMENU, Value.IMAGE).queue();
     }
 
+    public static void addNewTabFromContextMenu() {
+        TelemetryEvent.create(Category.ACTION, Method.ADD, Object.BROWSER_CONTEXTMENU, Value.LINK)
+                .extra(Extra.SOURCE, Object.BROWSER_CONTEXTMENU)
```
mTwTm: Duplicate Object.BROWSER_CONTEXTMENU?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -332,6 +338,10 @@ public static void copyImageEvent() {
         TelemetryEvent.create(Category.ACTION, Method.COPY, Object.BROWSER_CONTEXTMENU, Value.IMAGE).queue();
     }
 
+    public static void addNewTabFromContextMenu() {
+        TelemetryEvent.create(Category.ACTION, Method.ADD, Object.BROWSER_CONTEXTMENU, Value.LINK).queue();
```
mTwTm: It feels a bit weird to not see tab here but I'm ok with that.

PR #1748, opened 2018-04-02
build: add RxJava dependency

walkingice:
To involve RxJava to code base

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -213,6 +213,10 @@ dependencies {
     implementation Libs.adjust
     implementation Libs.play_services_analytics // Required by Adjust
 
+    // https://github.com/ReactiveX/RxAndroid
+    implementation 'io.reactivex.rxjava2:rxandroid:2.0.2'
+    implementation 'io.reactivex.rxjava2:rxjava:2.1.12'
+
     api Libs.lottie
 
     testImplementation Libs.junit


walkingice:
Per my build test

the release apk size is
**2371662** - before adding dependency
**2402586** - after adding dependency includes below changes

```java
Observable.just("one", "two", "three", "four", "five")
         .subscribeOn(Schedulers.newThread())
         .observeOn(AndroidSchedulers.mainThread())
          .subscribe(new Consumer<String>() {
             @Override
             public void accept(String s) throws Exception {
                 Log.d("TAG", s);
              }
         });
```

idiotmax:
How about let's make MainActivity.asyncInitialize as a real use case?

app/build.gradle
```diff
@@ -213,6 +213,10 @@ dependencies {
     implementation Libs.adjust
     implementation Libs.play_services_analytics // Required by Adjust
 
+    // https://github.com/ReactiveX/RxAndroid
+    implementation 'io.reactivex.rxjava2:rxandroid:2.0.2'
```
benjamin-cheng: Could you please help to define the dependency in the file below?
https://github.com/mozilla-tw/Rocket/blob/master/buildSrc/src/main/java/Dependencies.kt

Then you could reference here as:
`implementation Libs.rxandroid`
`implementation Libs.rxjava`

walkingice:
We had kotlin, and we don't have plan for RxJava for now.

PR #1762, opened 2018-04-03
Fix strict mode violations in screenshot viewer activity

benjamin-cheng:
Move File.exists(), File.delete() and BitmapFactory.decodeFile() to background thread

### DIFF
--- a/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotViewerActivity.java
+++ b/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotViewerActivity.java
@@ -17,6 +17,7 @@ import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.graphics.BitmapFactory;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotViewerActivity.java
```diff
@@ -453,13 +421,18 @@ public boolean onLongClick(View v) {
...
+                    if (file.exists()) {
+                        try {
+                            file.delete();
+                        } catch (Exception ex) {
+                        }
+                    }
                 }
-            }
+            }).start();
             ScreenshotManager.getInstance().delete(mScreenshot.getId(), this);
```
idiotmax: How about deleting the file in the callback?

benjamin-cheng: It seems fine to delete the file in `onDeleteComplete()`. However, I prefer to keep it here in this commit to focus on fixing the strict mode issue. What's the benefit you think to delete the file in the callback?

PR #1766, opened 2018-04-09
Nevin/issues/1084

### LINKED ISSUE #1084: Integrate Firebase [for Nightly channel, closed alpha testing]

cnevinc:
Add Firebase to Rocket includes Fix #1084

Analytics
Remote Config
Cloud Messaging
Crashlytics

For collaboration:
Although this approach is different with Firebase's document [1], it works better for us cause we don't want developers to copy server-side configurations to their sandbox projects.

External contributors can play with their sandbox projects and update google-service.json accordingly. But there's no guarantee that they'll work as expected ( should fall back to default behaviour if implemented correctly.)

[1] Large team, safer development


app/build.gradle
```diff
@@ -33,6 +33,7 @@ android {
         versionCode Versions.version_code
         versionName Versions.version_name
         testInstrumentationRunner "org.mozilla.focus.test.runner.CustomTestRunner"
+        testInstrumentationRunnerArgument 'disableAnalytics','true'
```
benjamin-cheng: 💯 Look like we could remove the CustomTestRunner after this flag turns on.

app/build.gradle
```diff
@@ -589,3 +592,5 @@ task checkstyle(type: Checkstyle) {
 afterEvaluate {
     check.dependsOn 'findbugs', 'pmd', 'checkstyle'
 }
+
+apply plugin: 'com.google.gms.google-services'
```
benjamin-cheng: Would you consider to move this `apply plugin ...` to the beginning of the `build.gradle`? It seems a conversion to have all the `apply plugin` there.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Provide helper for Firebase functionality
+ */
+public class FirebaseHelper {
+
+    public static final String RATE_APP_DIALOG_TEXT_TITLE = "rate_app_dialog_text_title";
+    public static final String RATE_APP_DIALOG_TEXT_CONTENT = "rate_app_dialog_text_content";
+    private static final String TAG = "FirebaseHelper";
+
+    private static WeakReference<FirebaseRemoteConfig> REMOTE_CONFIG;
```
benjamin-cheng: Variable with all capital letters is constant by conventions.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+    private static long CACHE_EXPIRATION = 3600; // 1 hour in seconds.
```
benjamin-cheng: should add `final` for it

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+        if (config != null) {
+            return config.getString(key);
+        }
+        return null;
+    }
+
+    public static void init(final Context context) {
+        new Thread(new Runnable() {
+            public void run() {
+                internalInit(context);
```
benjamin-cheng: Why put it to worker thread? Could there be issues to initial firebase in non-UI thread? For example, what if we access the remote config object in UI thread, however, the initializing code is not executed yet?

build.gradle
```diff
@@ -4,6 +4,10 @@ buildscript {
     repositories {
         jcenter()
         google()
+        maven { url "https://maven.google.com" }
```
benjamin-cheng: it's duplicated with the `google()` above.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+ * Provide helper for Firebase functionality
+ */
+final public class FirebaseHelper {
+
+    static final String RATE_APP_DIALOG_TEXT_TITLE = "rate_app_dialog_text_title";
+    static final String RATE_APP_DIALOG_TEXT_CONTENT = "rate_app_dialog_text_content";
+
+    private static final String TAG = "FirebaseHelper";
+
+    private static WeakReference<FirebaseRemoteConfig> REMOTE_CONFIG;
```
benjamin-cheng: Variable with all capital letters should be constant by conventions.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+    private static long CACHE_EXPIRATION = 3600; // 1 hour in seconds.
```
benjamin-cheng: Suggest to rename it to DEFAULT_CACHE_EXPIRATION_IN_SECONDS and add final for it. Then define another member variable cacheExpirationInSecond to store the exact expired time.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            return config.getString(key);
+        }
+        return null;
+    }
+
+    public static void init(final Context context) {
+        // internalInit() require I/O so I put it in background thread.
+        new Thread(new Runnable() {
+            public void run() {
+                internalInit(context);
```
benjamin-cheng: Is it necessary to initialize firebase in the worker thread? Could there be issues to initialize it in non-UI thread? For example, what if we access the remote config object in UI thread, however, the initializing code is not executed yet?

I found the initialize code is executed in UI thread in google sample code:
https://github.com/firebase/quickstart-android/blob/1b310e0e0ea2279d7e1363edb7e4923d0fc55369/config/app/src/main/java/com/google/samples/quickstart/config/MainActivity.java#L66

build.gradle
```diff
@@ -12,6 +16,8 @@ buildscript {
         classpath Libs.kotlin_gradle_plugin
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
+        classpath 'com.google.gms:google-services:3.1.1' // google-services plugin
```
benjamin-cheng: Please define these two dependencies in `Dependencies.kt`

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
cnevinc: I got a Strict Mode violation warning if initialize it on the main thread. So I guess it's required.

app/src/main/java/org/mozilla/focus/notification/RocketMessagingService.java
```diff
@@ -0,0 +1,43 @@
...
+import android.content.Intent;
+import android.net.Uri;
+import android.support.v4.app.NotificationCompat;
+
+import org.mozilla.focus.activity.MainActivity;
+
+public class RocketMessagingService extends FirebaseMessagingServiceWrapper {
+
+    @Override
+    public void onRemoteMessage(String url, String title, String body) {
```
cnevinc: onMessageReceived is implemented in FirebaseMessagingServiceWrapper ( if it's actual implementation).
And onRemoteMessage() will be called in there.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -1,97 +1,40 @@
...
 import java.util.HashMap;
 
 /**
- * Provide helper for Firebase functionality
+ * Implementation for FirebaseWrapper.
+ * Also need to provide implementation of getRemoteConfigDefault() method
  */
-final public class FirebaseHelper {
-
+final public class FirebaseHelper extends FirebaseWrapper {
```
cnevinc: Helper will be just an implementation to the wrapper 
1. start the wrapper (in a background thread)
2. implement getRemoteConfigDefault to .... provide Remote Config Default

app/build.gradle
```diff
@@ -219,10 +221,7 @@ dependencies {
 
     implementation Libs.adjust
     implementation Libs.play_services_analytics // Required by Adjust
-    implementation Libs.firebase_core
```
cnevinc: move all the dependencies to the module

cnevinc:
Please use https://github.com/mozilla-tw/Rocket/pull/1844 to review

PR #1769, opened 2018-04-09
Issue 1703 default notification once

walkingice:
If user already set Rocket as default browser, we don't show notification

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/Browsers.java
+++ b/app/src/main/java/org/mozilla/focus/utils/Browsers.java
@@ -214,9 +214,10 @@ public class Browsers {
     /**
      * Is *this* application the default browser?
      */
-    public boolean isDefaultBrowser(Context context) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -214,9 +214,10 @@ public boolean isInstalled(KnownBrowser browser) {
     /**
      * Is *this* application the default browser?
      */
-    public boolean isDefaultBrowser(Context context) {
-        return defaultBrowser != null && context.getPackageName().equals(defaultBrowser.packageName);
-
+    public static boolean isDefaultBrowser(Context context) {
+        final Browsers browser = new Browsers(context, "http://mozilla.org");
```
cnevinc: Can we make this browser instance a singleton so we don't have to create it multiple times?

walkingice: If it worth to do then we can file a new issue to do it

PR #1778, opened 2018-04-10
Show smile face ':)' on tab counter before the tabs are restored

### LINKED ISSUE #1699: Often see tab number showing 1 and then jump to the correct number when opening link from gmail (sometimes the lag is long)
(Issued by Joe)

STR:
 1. Launch Rocket without any tab
 2. Add a new tab 
 3. Open a new link from an external app (Gmail)
 4. See tab number

Actual result:
 - Often see tab number showing 1 and then jump to the correct number when opening a link from Gmail (sometimes the lag is long)

Device: Sony Z3C
Android: 6.0
Rocket: 1.0.4(3132).nightly

benjamin-cheng:
Already reviewed by the designers. Close #1699 

rocket-smile-tab-counter-shift-one-dp

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -58,6 +58,7 @@ import android.widget.TextView;
 import android.widget.Toast;
 
 import org.mozilla.focus.R;
+import org.mozilla.focus.activity.MainActivity;
 import org.mozilla.focus.activity.ScreenNavigator;
 import org.mozilla.focus.download.DownloadInfo;
 import org.mozilla.focus.download.DownloadInfoManager;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -1050,6 +1051,11 @@ private boolean isPopupWindowAllowed() {
         return ScreenNavigator.get(getContext()).isBrowserInForeground();
     }
 
+    private boolean isTabRestoredComplete() {
+        return (getActivity() instanceof MainActivity)
```
mTwTm: NIT: throw an exception if not instanceof MainActivity so we won't have a false that does not really mean tab is restoring?

benjamin-cheng: Good idea! Refined.

PR #1784, opened 2018-04-11
Issues/1783

### LINKED ISSUE #1783: NullPointer exception due to getFocusTab() and getTabView()

mTwTm:
Fixes #1783


mTwTm:
for case 1:
Oops I missed that one, I'll add it.

for case 2:
The tabview should be right created after addTabInternal() and there should be a tabview.

mTwTm:
@pharspher I'm uncertain with the 
```
                // This should not happen, but if once happened re-Focus again to force initialization.
                tabsSession.switchToTab(targetTab.getId());
```
part. Does that make sense?

PR #1790, opened 2018-04-11
[TabTray] Workaround to avoid crash when closing a tab

pharspher:
This workaround solution will not sync tab tray UI with tab data, i.e. The closed tab will still be visible on tab tray.

Result behavior:
Click the closed item => close tab tray, and show the previous page.
Dismiss the closed item => remove the item, and switch to the next focused page.


idiotmax:
I think we have the full solution rather than this workaround solution merged. I'll close this one, reopen if you have other concern? @pharspher  @walkingice

PR #1803, opened 2018-04-14
Add leankanary for debug build

cnevinc:
Not sure if we've discussed it before. Maybe we can add Leankanary for easier finding leaks in debug build.


idiotmax:
I believe this is duplicated with #1736, and the patch is almost the same XD.

PR #1806, opened 2018-04-17
Fix strict mode issues for downloading 

benjamin-cheng:


app/src/main/java/org/mozilla/focus/download/EnqueueDownloadTask.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Use Android's Download Manager to queue this download.
+ */
+public class EnqueueDownloadTask extends AsyncTask<Void, Void, EnqueueDownloadTask.ErrorCode> {
+
+    private WeakReference<Activity> activityRef;
+    private Download download;
+    private String refererUrl;
+
+    public EnqueueDownloadTask(@NonNull Activity activity, @NonNull Download download, @Nullable String refererUrl) {
```
idiotmax: Do we really need an activity? or just a context to provide the localized string from resources?

Why are we not passing these constructor parameters to execute(context, download, string)?

app/src/main/java/org/mozilla/focus/download/DownloadInfoManager.java
```diff
@@ -101,21 +101,31 @@ protected void onUpdateComplete(int token, Object cookie, int result) {
...
-                    if (cookie != null) {
-                        List<DownloadInfo> downloadInfoList = new ArrayList<>();
-                        if (cursor != null) {
-                            while (cursor.moveToNext()) {
-                                final DownloadInfo downloadInfo = cursorToDownloadInfo(cursor);
-                                downloadInfoList.add(downloadInfo);
+                    ThreadUtils.postToBackgroundThread(new Runnable() {
+                        @Override
+                        public void run() {
+                            if (cookie != null) {
```
idiotmax: We didn't close the cursor when the cookie is null, right?

app/src/main/java/org/mozilla/focus/download/DownloadInfoManager.java
```diff
@@ -101,21 +101,31 @@ protected void onUpdateComplete(int token, Object cookie, int result) {
...
+                                final List<DownloadInfo> downloadInfoList = new ArrayList<>();
+                                if (cursor != null) {
+                                    while (cursor.moveToNext()) {
+                                        final DownloadInfo downloadInfo = cursorToDownloadInfo(cursor);
+                                        downloadInfoList.add(downloadInfo);
+                                    }
+                                    cursor.close();
```
idiotmax: There's a pattern when dealing with a cursor, consider use try/catch/finally to handle the cursor lifecycle.

app/src/main/java/org/mozilla/focus/download/EnqueueDownloadTask.java
benjamin-cheng: The context are used in both `doInBackground` and `onPostExecute`. Besides, I would like it's an activity object to have some life-cycle checking (if user left the activity) at the beginning of the `onPostExecute`.

app/src/main/java/org/mozilla/focus/download/DownloadInfoManager.java
benjamin-cheng: Fixed. Thanks!

PR #1810, opened 2018-04-18
Move PermissionHandler to a standalone module

mTwTm:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -91,6 +89,8 @@
 import org.mozilla.focus.widget.BackKeyHandleable;
 import org.mozilla.focus.widget.FragmentListener;
 import org.mozilla.focus.widget.TabRestoreMonitor;
+import org.mozilla.rocket.permissionhandler.permission.PermissionHandle;
```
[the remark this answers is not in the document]
mTwTm: I'm thinking of part of an advertisement but if we're moving to mozilla-mobile then I think yes I should change it.

app/build.gradle
```diff
@@ -242,6 +242,7 @@ dependencies {
     betaImplementation Libs.leakcanary_no_op
     coverageImplementation Libs.leakcanary_no_op
     releaseImplementation Libs.leakcanary_no_op
+    implementation project(path: ':modules:permissionhandler')
```
benjamin-cheng: Suggest to move this definition to the beginning of `dependencies` section sitting with other project dependencies.

modules/permissionhandler/build.gradle
```diff
@@ -0,0 +1,35 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 27
```
benjamin-cheng: Should apply the `versions.gradle` settings.

app/src/main/res/values/strings.xml
```diff
@@ -195,7 +195,7 @@
     <string name="label_menu_settings">Settings</string>
     <string name="label_menu_exit">Exit</string>
 
-    <string name="permission_dialog_setting">settings</string>
+    <string name="permission_handler_permission_dialog_setting">settings</string>
```
benjamin-cheng: Should we remove this string definition since it's already defined in `modules/permissionhandler/src/main/res/values/strings.xml`?

mTwTm: We want to keep it so it'll be localized by the l10n team

mTwTm:
@benjamin-cheng updated :)

mTwTm:
Close Favoring #1856

PR #1816, opened 2018-04-19
Enlarge the touch area of home screen's menu buttons. Close #1502

### LINKED ISSUE #1502: The clickable area of 'tab tray' icon on home feels too small
(Issued by Carol)

The clickable area of 'tab tray' icon on home feels too small

Device: GooglePixel
Rocket nightly: 2.0.0 (2383)
Android version: 8.1.0

benjamin-cheng:
device-2018-04-19-175754

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -189,4 +189,9 @@
         <item name="android:background">@drawable/round_rectangle_ripple</item>
     </style>
 
+    <style name="HomeMenuButton" parent="MainMenuButton">
+        <item name="android:layout_height">48dp</item>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -189,4 +189,10 @@
         <item name="android:background">@drawable/round_rectangle_ripple</item>
     </style>
 
+    <style name="HomeMenuButton">
```
mTwTm: NIT: how about extending this from MainMenuButton and only change the size?

benjamin-cheng: OK. Updated!

PR #1817, opened 2018-04-19
Library and plugin upgrades.

mTwTm:


app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.java
```diff
@@ -325,6 +325,11 @@ private Drawable buildDrawable(@NonNull Drawable original,
         }
     }
 
+    @SuppressWarnings("ResourceType")
+    private boolean isValidInterpolator(@InterpolatorRes int itplId) {
+        return itplId > 0;
```
idiotmax: Since the android lint tells us that using ```> 0``` to validate the resource is incorrect, why do we suppress the lint here?

mTwTm: My understanding to this is the lint is trying to avoid the situation when a resource should be used by getting the value is misused by using the id directly. However in this case, I believe the dev of this block is trying to use the resource directly to make sure there is such resource. ( >0 )

app/build.gradle
```diff
@@ -188,60 +188,60 @@ dependencies {
     implementation project(':third_party:subsampling-scale-image-view')
     implementation project(':third_party:glide:annotation')
     implementation project(':third_party:glide:library')
-    annotationProcessor Libs.glide_compiler
-    kapt Libs.glide_compiler
+    annotationProcessor "com.github.bumptech.glide:compiler:${Versions.glide}"
```
benjamin-cheng: Should we merge the #1808 first then rebase it?

mTwTm: I think I'll do the reverse, I'll rebase 1808 since rebasing that one is simpler :P

app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.java
mTwTm: @walkingice Can you share with us what's the rationale behind itplId > 0? Is it a sanity check as I've described?

walkingice: yes it is a checking. But I am not sure whether it is a good practice.

mTwTm: Fixed :)

PR #1825, opened 2018-04-20
Remove android text resolver dependency

idiotmax:
Android Text Resolver is a gradle plugin to allow us replace a string inside another string by placeholder, ex.```{{app_name}}```. The necessary use case is the string used in ```AndroidManifest.xml``` only allows plain text, not formatted text, ex. ```%1$s```. However, this plugin is broken after ```aapt2``` enabled with 3.1.0 android gradle plugin, and output the merged resource file into ```.flat``` binary format, cause the string not able to be resolved by the plugin.

Thus, this plugin is our SDK upgrade blocker.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
@@ -5,10 +5,13 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.support.v4.app.NotificationCompat;
+import android.support.v4.content.ContextCompat;
+import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.app.AlertDialog;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.TextView;
 
 import org.mozilla.focus.R;
 import org.mozilla.focus.activity.MainActivity;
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
```diff
@@ -43,6 +46,10 @@ public void onCancel(DialogInterface dialogInterface) {
         });
 
         View dialogView = LayoutInflater.from(context).inflate(R.layout.layout_rate_app_dialog, (ViewGroup) null);
+
+        final TextView textView = dialogView.findViewById(R.id.rate_app_dialog_textview_title);
+        textView.setText(context.getString(R.string.rate_app_dialog_text_title, context.getString(R.string.app_name)));
```
mTwTm: NIT: remove the text in layout file so devs won't attempt to change these text there?

app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
```diff
@@ -105,6 +112,10 @@ public void onCancel(DialogInterface dialogInterface) {
         });
 
         View dialogView = LayoutInflater.from(context).inflate(R.layout.layout_share_app_dialog, (ViewGroup) null);
+
+        final TextView textView = dialogView.findViewById(R.id.share_app_dialog_textview_title);
+        textView.setText(context.getString(R.string.share_app_dialog_text_title, context.getString(R.string.app_name)));
```
mTwTm: NIT: remove the text in layout file so devs won't attempt to change these text there?

gradle.properties
```diff
@@ -16,4 +16,4 @@ org.gradle.jvmargs=-Xmx1536m
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
-android.enableAapt2=false
+#android.enableAapt2=false
```
benjamin-cheng: Just suggestion, we could remove it instead of comment out.

PR #1826, opened 2018-04-20
Convert png to webp to save apk size by 53kb

mxlius:


benjamin-cheng:
There is a downside to convert png to webp that the file cannot be previewed by finder or github. In this case, I cannot see the exact changes you made in the pull request. So, I am used to only convert high resolution png image into webp since it may not save much bytes for small images.

mTwTm:
bitrot

PR #1827, opened 2018-04-22
Make fields final when possible.

gaul:
Found via error-prone 2.3.1 FieldCanBeFinal check.


mTwTm:
Thanks for the contribution! However this pr opened up a discussion in our dev team around coding style and process. We're currently (slowly) debating if an enforced final operator is helping the overall code quality since sometimes a coder may explicitly leave variables as non-final while it may be only assigned once at that time. A static check may fail to differentiate these situations and it is also possible that a patcher and reviewer is uncertain of these situations too. So this patch may wait for a while since we're still discussing how we'll deal with these.

gaul:
Closing due to bitrot.

Issue #1412, opened 2018-02-09
For update users, prompt users after they open the updated app with Firebase


joechengla:

weslyhuang:
notify users about Firebase. next step on UX for a design proposal.

Jacklin-ux:
Spec link:
https://drive.google.com/open?id=1zNaVfTurjgP1Rc49CUOhdHfSKvFzXIEk

Question: After users update Firefox Rocket from 1.X to 2.1, open Rocket then show onboarding page with "Privacy policy update" notification. I am wondering if users press the notification at that moment, can we close onboarding page then open "Privacy policy update" page in the new tab?

cnevinc:
1. Making notifications' importance to "LOW" will make 8.0+ devices won't be able to show heads-up.
2. Is this issue here only for the app update? I'm imaging the workflow below
a. An existing user opens Rocket for the first time after the update.
b. The app should display a notification of the privacy notice
c. if the user clicks the notification, and at that moment the onboarding page is showing, we should hide the onboarding page and create a new tab for "Privacy policy update" page.

My question will be
1. In spec page2, I don't know what "Push" means. There's nothing to push from GCM here.
2. for a, so new users won't see this notification?
3. for b, While the user launches the app, if the user has seen previous onboarding(2.0), but haven't' finish the onboarding for the new version(2.1), I'll treat it as an updated user. I can display the notification at this moment. But are we showing a new onboarding for 2.1? 
4. for c, does the user will see the onboarding again ?

cnevinc:
Just checked with Jack
1. Only "updated users" ( e.g. 1.1->2.1, 2.0->2.1) need to see this notification.
2. For 2.1 users who will see the onboarding, won't see this notification.
3. Priority/importance for this notification is High (for heads-up)

I still need spec for:
What will happen if the user click on the notification? Open a SUMO page or local static html?

weslyhuang:
In for May. 12.
1. still need UX to notify updated user. 
2. may need SUMO for it too.
3. two strings added.

chianger:
@benjamin-cheng latest changes on the strings
Title: Product Update Notice
Description:
To make Firefox Rocket better for you and everyone, we made some changes to our privacy policy. Tap to learn more.

weslyhuang:
@benjamin-cheng just to be sure, are we still lacking strings and spec for this issue? or we can remove the "UX&VD Wanted" and "String Wanted" labels now? thanks.

PR #1834, opened 2018-04-23
Add strings for privacy policy update notification (#1412)

benjamin-cheng:


app/src/main/res/values/strings.xml
```diff
@@ -312,6 +312,10 @@
     <!-- Remove Top site item -->
     <string name="remove">Remove</string>
 
+    <!-- In-app notification, Privacy policy update -->
+    <string name="privacy_policy_update_notification_title">Privacy policy update</string>
+    <string name="privacy_policy_update_notification_action">Tap here to learn more about the changes.</string>
```
mTwTm: NIT: content seems more natural to me than action but you can decide.

PR #1844, opened 2018-04-25
Nevin/issues/1084 0424

### LINKED ISSUE #1084: Integrate Firebase [for Nightly channel, closed alpha testing]

cnevinc:
Add Firebase to Rocket includes Fix #1084

Analytics
Remote Config
Cloud Messaging
Crashlytics

For collaboration:
Although this approach is different with Firebase's document [1], it works better for us cause we don't want developers to copy server-side configurations to their sandbox projects.

External contributors can play with their sandbox projects and update google-service.json accordingly. But there's no guarantee that they'll work as expected ( should fall back to default behaviour if implemented correctly.)

[1] Large team, safer development


benjamin-cheng:
Should we move those `google-services.json` to CI server? Or anyone could see those `api_key` information in the files.

mTwTm:
Disclaimer: I know nothing about Firebase and these keys.
However if these keys are not supposed to be revealed, we might also want to renew them before production cause they're already leaked on Github in this pr?

PR #1856, opened 2018-04-26
Import permission handler as submodule

mTwTm:
I think importing with submodule is better than copying the whole source codes in #1810. With submodule we can:
1. Flexibly checkout specific version of the target library when we want to.
2. Prevent Rocket contributors to change the library implementation in the cloned version in Rocket.


mTwTm:
We'll need to do `git pull --recurse-submodules` after this commit. (Including future clones.)

PR #1863, opened 2018-04-30
Update strings for privacy policy update notification (#1412)

benjamin-cheng:


benjamin-cheng:
Update the strings only, so merge it directly.

PR #1886, opened 2018-05-03
Issue 346 improve set default browser

### LINKED ISSUE #346: improved set default browser
Improve the guidance to end users after prompting user to set Zerda as default browser.

walkingice:
This PR tries to improve set-default-browser to resolve #346, by

* Add a ConfigActivity, by enabling/disabling it, we can wipe default-browser-config
* Add a Service to enable/disable ConfigActivity in background
    * Service will turn off itself after 30 seconds as timeout
    * Service will run as foreground service
    * Service will post notification, which notification will launch SettingsActivity for user convenience
    * Service will notify Preference by sending broadcast, when the service is finished
* Preference will start service to wipe config, if sdk < N.


app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * A helper function to report whether this service is alive
+     *
+     * @param context
+     * @return true if this service is alive
+     */
+    public static boolean isAlive(Context context) {
+        final ActivityManager manager = (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
```
cnevinc: There'll be only one service at a time?

app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
```diff
@@ -0,0 +1,257 @@
...
+                new Intent(getApplicationContext(), SettingsActivity.class),
+                PendingIntent.FLAG_UPDATE_CURRENT);
+    }
+
+    // A TimerTask to stop service if it runs too long
+    private final static class BombTask extends TimerTask {
+        // How long this Service will stop itself
+        private static final long TIMEOUT = 30000; // 30 seconds
+
+        final ComponentToggleService service;
```
cnevinc: Maybe use a weakReference?

app/src/main/java/org/mozilla/focus/widget/DefaultBrowserPreference.java
```diff
@@ -75,9 +119,120 @@ private void openDefaultAppsSettings(Context context) {
...
+        }
+
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            // update UI
+            pref.update();
+
+            // SettingsActivity is in foreground(because this BroadcastReceiver is working),
+            // to remove notification which created by Service
+            NotificationManagerCompat.from(context).cancel(ComponentToggleService.NOTIFICATION_ID);
```
cnevinc: I'm wondering if the user pref off "Send Usage Data" and then Firebase was also stopped, what will happen here...

app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
walkingice: Yes, and it is a foreground service

app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
walkingice: The task itself will be canceled once the service be destroyed. To use WeakReference here won't hurt anything, so I updated it as your suggestion.

app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        return false;
+    }
+
+    private void startToForeground() {
+        //  The notification channel id should not be necessary since this service should only be enabled for API 21~23
+        final NotificationCompat.Builder builder = new NotificationCompat.Builder(
+                getApplicationContext(),
+                NOTIFICATION_CHANNEL);
```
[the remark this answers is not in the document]
walkingice: Channel Id does not matter to this notification since it only runs on old Android. I will reuse the existing one.

app/src/main/java/org/mozilla/focus/widget/DefaultBrowserPreference.java
walkingice: Nothing is supposed to change here

app/src/main/java/org/mozilla/focus/components/ComponentToggleService.java
```diff
@@ -49,6 +48,12 @@
     private static final int INTENT_REQ_CODE = 0x9527;
     private static final IntentFilter sIntentFilter = new IntentFilter();
 
+    /**
+     * This is foreground service, so we can ensure if the service be killed on low memory, this
+     * field will be reset as well.
+     */
+    private static boolean sIsAlive = false;
```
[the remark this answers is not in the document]
walkingice: To make writing test easier is a good point to me. I will revert the previous patch to get ActivityManager-query back.

Issue #1411, opened 2018-02-09
Allow users to turn off Firebase sharing in Settings


joechengla:
Let’s tie Firebase setting to Send Usage Data setting (global switch)

weslyhuang:
next step on UX for a design proposal.

cnevinc:
I'll need the UI design so I know where to add the setting.

cnevinc:
Disable Analytics
https://firebase.google.com/support/guides/disable-analytics
Disable Crashlytics (need to restart the app)
https://firebase.google.com/docs/crashlytics/customize-crash-reports#enable_opt_in_reporting
Disable the performance monitoring ( Not implemented yet)
https://firebase.google.com/docs/perf-mon/disable-sdk



[1] https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId.html#deleteInstanceId()

weslyhuang:
Need UX/Product detail. (Risky for May 12?)
1. new string
2. may need SUMO for it too.

weslyhuang:
@cnevinc just to be sure, are we still lacking strings and spec for this issue? or we can remove the "UX&VD Wanted" and "String Wanted" labels now? thanks.

PR #1894, opened 2018-05-04
Nevin/issues/1411

cnevinc:
Cloese #1411


cnevinc:
Placing FirebaseHelper.init() is better than Application onCreate
Since ContentProvider's onCreate will always call before Application's onCreate.
See: http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/java/android/app/ActivityThread.java#5386 and 
http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/java/android/app/ActivityThread.java#5405

cnevinc:
Here's the document for library support for Fabric
https://docs.fabric.io/android/crashlytics/build-tools.html#set-up-a-library-subproject

But the approach is changed since it now relies on google_app_id to generate its API key.
But setting up the build id part in the link is still useful.

PR #1904, opened 2018-05-08
 Prompt privacy policy update notification when app is updating from 2.0 to 2.1

### LINKED ISSUE #1412: For update users, prompt users after they open the updated app with Firebase

benjamin-cheng:
Close #1412

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/utils/AndroidTestUtils.java
+++ b/app/src/androidTest/java/org/mozilla/focus/utils/AndroidTestUtils.java
@@ -22,7 +22,6 @@ import org.mozilla.focus.BuildConfig;
 import org.mozilla.focus.Inject;
 import org.mozilla.focus.R;
 import org.mozilla.focus.activity.MainActivity;
-import org.mozilla.focus.fragment.FirstrunFragment;
 import org.mozilla.focus.widget.FragmentListener;
 
 import java.io.IOException;
@@ -34,8 +33,6 @@ import okio.Okio;
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.action.ViewActions.click;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
-import static org.mozilla.focus.fragment.FirstrunFragment.PREF_KEY_BOOLEAN_FIRSTRUN_SHOWN;
-import static org.mozilla.focus.fragment.FirstrunFragment.PREF_KEY_INT_FIRSTRUN_UPGRADE_VERSION;
 import static org.mozilla.focus.utils.RecyclerViewTestUtils.clickChildViewWithId;
 
 public final class AndroidTestUtils {
@@ -44,19 +41,18 @@ public final class AndroidTestUtils {
         beforeTest(true);
     }
 
-    public static void beforeTest(final boolean firstRun) {
+    public static void beforeTest(final boolean skipFirstRun) {
         final Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
         if (context == null) {
             return;
         }
-        final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-        if (preferences != null) {
-            final SharedPreferences.Editor editor = preferences.edit();
-            if (editor != null) {
-                editor.putBoolean(PREF_KEY_BOOLEAN_FIRSTRUN_SHOWN, firstRun)
-                        .putInt(PREF_KEY_INT_FIRSTRUN_UPGRADE_VERSION, FirstrunFragment.FIRSTRUN_UPGRADE_VERSION).commit();
-            }
+
+        if (skipFirstRun) {
+            NewFeatureNotice.getInstance(context).setMultiTabUpdateNoticeDidShow();
+        } else {
+            NewFeatureNotice.getInstance(context).resetFirstRunDidShow();
         }
+
         final Settings settings = Settings.getInstance(context);
         if (settings != null) {
             settings.setShareAppDialogDidShow();
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -65,6 +65,7 @@ import org.mozilla.focus.utils.FileUtils;
 import org.mozilla.focus.utils.FirebaseHelper;
 import org.mozilla.focus.utils.FormatUtils;
 import org.mozilla.focus.utils.IntentUtils;
+import org.mozilla.focus.utils.NewFeatureNotice;
 import org.mozilla.focus.utils.NoRemovableStorageException;
 import org.mozilla.focus.utils.SafeIntent;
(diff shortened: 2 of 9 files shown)


mTwTm:
I'll takeover this review since Nevin is super busy.

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public void setMultiTabUpdateNoticeDidShow() {
+        preferences.edit()
+                .putBoolean(PREF_KEY_BOOLEAN_FIRSTRUN_SHOWN, true)
+                .putInt(PREF_KEY_INT_FEATURE_UPGRADE_VERSION, MULTI_TAB_FROM_VERSION_1_0_TO_2_0)
+                .apply();
+    }
+
+    public boolean shouldShowPrivacyPolicyUpdate() {
```
mTwTm: I don't think this will ever be true.

app/src/androidTest/java/org/mozilla/focus/utils/AndroidTestUtils.java
```diff
@@ -49,14 +46,9 @@ public static void beforeTest(final boolean firstRun) {
...
-        final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
-        if (preferences != null) {
-            final SharedPreferences.Editor editor = preferences.edit();
-            if (editor != null) {
-                editor.putBoolean(PREF_KEY_BOOLEAN_FIRSTRUN_SHOWN, firstRun)
-                        .putInt(PREF_KEY_INT_FIRSTRUN_UPGRADE_VERSION, FirstrunFragment.FIRSTRUN_UPGRADE_VERSION).commit();
-            }
-        }
+
+        NewFeatureNotice.getInstance(context).setFirstRunDidShow();
```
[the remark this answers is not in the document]
benjamin-cheng: Done

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
benjamin-cheng: My bad. I forgot to flip it back after testing. Also having some minor modifications for the update path - 1.0.x to 2.1. Please help to review, thank you.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -333,6 +334,10 @@ private void runPromotion(final SafeIntent intent) {
             postSurveyNotification();
             history.add(Settings.Event.PostSurveyNotification);
         }
+
+        if (NewFeatureNotice.getInstance(this).shouldShowPrivacyPolicyUpdate()) {
```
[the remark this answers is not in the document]
benjamin-cheng: It should not be problem since the privacy policy notification would be shown with others. User still can tap it in the notification list.

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
```diff
@@ -0,0 +1,87 @@
...
+    @VisibleForTesting
+    public void resetFirstRunDidShow() {
+        preferences.edit()
+                .putBoolean(PREF_KEY_BOOLEAN_FIRSTRUN_SHOWN, false)
+                .putInt(PREF_KEY_INT_FEATURE_UPGRADE_VERSION, 0)
+                .apply();
+    }
+
+    private boolean isNewlyInstalled() {
+        return !hasShownFirstRun() && (getLastShownFeatureVersion() == 0);
```
mTwTm: I think !hasShownFirstRun() is enough but if we're planning to deprecate hasShownFirstRun() with versionNum then we should also keep (getLastShownFeatureVersion() == 0) here

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
```diff
@@ -0,0 +1,87 @@
...
+        setLastShownFeatureVersion(MULTI_TAB_FROM_VERSION_1_0_TO_2_0);
+    }
+
+    public boolean shouldShowPrivacyPolicyUpdate() {
+        if (isNewlyInstalled()) {
+            setPrivacyPolicyUpdateNoticeDidShow();
+            return false;
+        }
+
+        return FIREBASE_FROM_VERSION_2_0_TO_2_1 == getLastShownFeatureVersion() + 1;
```
mTwTm: maybe just return true here?

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void setPrivacyPolicyUpdateNoticeDidShow() {
+        setLastShownFeatureVersion(FIREBASE_FROM_VERSION_2_0_TO_2_1);
```
mTwTm: I think this may be override to v1 if the user first tap the privacy notice but later come back to finish 1.0 -> 2.0 onboarding?

mTwTm: I think adding an extra privacy notice shared preference would be simpler.

benjamin-cheng: 1. That won't happen. The notification would only show after the 1.0 => 2.0 onboarding is done protected by the statement below in `shouldShowPrivacyPolicyUpdate()`
`return FIREBASE_FROM_VERSION_2_0_TO_2_1 == getLastShownFeatureVersion() + 1;`

2. Yes, it would be more simple to code that way, however, we need to keep adding new preference for the new feature hint each time in the future. The preference could finally blow up.  So. I try reusing the same preference for the app upgrade kind task.

mTwTm: I just tried again and I noticed after onboarding is shown we'll receive the notice. I see why you're coding these this way now.

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
benjamin-cheng: I explained the statement in the thread below.

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
mTwTm: I'm thinking of privacy notice as not a version of `NewFeatureNotice` since it's a Notification rather then a new Onboarding content, so I think it is this nature of different displays that make me feel like separating them could be simpler. (Notification v1, Onboarding v2...etc) But current implementation is better in a way that we only display one thing at a time by forcing user to upgrade once at a time. (No simultaneous Notification and Onboarding) I'm just a bit worried about since getting more and more complex when we're having more upgrade versions and some of them should always be shown, while some of them should be skipped. These protection need to be handled very carefully so the old (v1, v2 when v3 is coming) messages will work correctly.

PR #1911, opened 2018-05-10
Nevin/issues/1411 0510

cnevinc:


app/src/androidTest/java/org/mozilla/focus/activity/FirebaseSwitcherTest.java
```diff
@@ -0,0 +1,244 @@
...
+        // Now launch Rocket's setting activity
+        settingsActivity.launchActivity(new Intent());
+
+//        FirebaseHelper.init(settingsActivity.getActivity());
+
+        FirebaseHelper.injectEnablerCallback(new Delay());
+
+        // Click on the switch multiple times...
+        return onData(
+                is(instanceOf(TelemetrySwitchPreference.class))).
```
mxlius: Can we find the preference by its key ```R.string.pref_key_default_browser``` instead of knowing the implementation class?

app/src/firebase/java/org/mozilla/focus/utils/AdjustHelper.java
```diff
@@ -0,0 +1,14 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.utils;
+
+import android.content.Context;
+
+public class AdjustHelper {
```
mxlius: Firebase should not know about Adjust, right?

app/src/main/AndroidManifest.xml
```diff
@@ -180,11 +180,13 @@
             </intent-filter>
         </receiver>
 
-
-        <service
```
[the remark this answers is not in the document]
cnevinc: typo. I'll add it back

app/src/main/AndroidManifest.xml
```diff
@@ -180,11 +180,13 @@
...
-            android:name="org.mozilla.focus.components.RelocateService"
+        <provider
+            android:name=".provider.FirebaseHelperProvider"
```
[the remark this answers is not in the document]
cnevinc: fixed.

app/src/main/AndroidManifest.xml
```diff
@@ -119,9 +119,9 @@
 
         <!-- We don't need Crashlytics in debug build. Besides, there's some Strict Mode violation
```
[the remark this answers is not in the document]
cnevinc: i'll remove it

app/src/firebase/java/org/mozilla/focus/utils/AdjustHelper.java
cnevinc: No. But a dummy implementation is still needed for firebase build type.

app/src/androidTest/java/org/mozilla/focus/activity/FirebaseSwitcherTest.java
cnevinc: TelemetrySwitchPreference's persistent is false. It has no key associated with it. I guess it's not possible to withKey() method to find it

app/src/main/java/org/mozilla/focus/FocusApplication.java
```diff
@@ -30,10 +32,10 @@ public void onCreate() {
         }
         LeakCanary.install(this);
 
-        PreferenceManager.setDefaultValues(this, R.xml.settings, false);
-
         // Provide different strict mode penalty for ui testing and production code
-        Inject.enableStrictMode();
+        enableStrictMode();
+
+        PreferenceManager.setDefaultValues(this, R.xml.settings, false);
```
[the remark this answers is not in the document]
cnevinc: Nice catch. Looking at the document, it said: " This should be called by the application's main activity."[1]
And we need the pref even before that, so I think putting it here is good enough.
I even think this should put in our FirebaseHelperProvider (and set CP's android:initOrder). Or even move` FirebaseHelper.init()` after `PreferenceManager.setDefaultValues(this, R.xml.settings, false);` in FocusApplication.

[1] https://developer.android.com/reference/android/preference/PreferenceManager#setdefaultvalues_3

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -13,32 +23,188 @@
...
+    public static boolean bind(@NonNull final Context context) {
+
+        final boolean enable = TelemetryWrapper.isTelemetryEnabled(context);
+
+        return enableFirebase(context.getApplicationContext(), enable);
+
+
+    }
+
+    // return true if a new runnable is created. otherwise return false.
```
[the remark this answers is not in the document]
cnevinc: OK!

app/src/main/AndroidManifest.xml
```diff
@@ -117,12 +117,6 @@
             android:exported="false"
             tools:ignore="ExportedService" />
 
-        <!-- We don't need Crashlytics in debug build. Besides, there's some Strict Mode violation
```
cnevinc: now Crashlytics is started by our code. So this is not needed.

cnevinc:
Just for the record
I've found some GC related test  [1][2][3][4][5] But I'm not sure if the tests will ends in 60 sec( to avoid the default time threshold for idlingResource... or maybe those examples are giving enough us confidence since some of them are run as unit tests?) 

The gc code I've copied can be found  in ZygoteInit [6] and Leakcanary is looking at [7]

[1] http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/tests/coretests/src/android/content/pm/PackageManagerTests.java#1319
[2] http://androidxref.com/7.1.2_r36/xref/libcore/benchmarks/src/benchmarks/ReferenceBenchmark.java#11
[3] http://androidxref.com/7.1.2_r36/xref/libcore/support/src/test/java/libcore/java/lang/ref/FinalizationTester.java#28
[4] http://androidxref.com/7.1.2_r36/xref/frameworks/base/tests/CoreTests/android/core/HeapTest.java#105
[5] http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/java/android/test/InstrumentationTestCase.java
[6] http://androidxref.com/7.1.2_r36/xref/frameworks/base/core/java/com/android/internal/os/ZygoteInit.java#494
[7] https://github.com/square/leakcanary/blob/master/leakcanary-watcher/src/main/java/com/squareup/leakcanary/GcTrigger.java#L31

app/src/main/java/org/mozilla/focus/Inject.java
```diff
@@ -34,14 +53,10 @@ public static void enableStrictMode() {
         final StrictMode.ThreadPolicy.Builder threadPolicyBuilder = new StrictMode.ThreadPolicy.Builder().detectAll();
         final StrictMode.VmPolicy.Builder vmPolicyBuilder = new StrictMode.VmPolicy.Builder().detectAll();
 
-        if (AppConstants.isBetaBuild()) {
```
mxlius: We don't check the build type and always popup dialog with log?

cnevinc: if it's Release build, we'll return immediately.
if it's Beta / debug / firebase build, we'll always popup dialogue with log
in AndroidTest, we only log error

PR #1919, opened 2018-05-15
Gradle refinements

mTwTm:
(WIP, planned in 2.2 so won't merge before 2.1 launched.)
Currently covered:
Move some part of build.gradle to a standalone file.
Planned:
Build type and flavor refinement.
Better dependency versioning.


mTwTm:
These are some items that I'd like to land first. I'm uncertain if we would like to remove coverage and firebase build-types but I think I can first remove the beta build type.

PR #1945, opened 2018-05-17
Closes #1943: Mock final class AppConstants (WIP)

### LINKED ISSUE #1943: Refine FirebaseEventTest
Mock AppConstants.isReleaseBuild()  to always return false so in app:testFocusWebkitReleaseUnitTest  it'll pass.
I though Unit Tests only runs on debug build type. just like Espresso. Also need to find out why.

cnevinc:
Closes #1943 
Use PowerMock to mock static and final class.
After I change all Mockito.mock to PowerMock.mock, everything works.

I need to understand how they work and what's the difference before it's landed.

I've also found that we run unit test on release build on purpose.  (find 'testFocusWebkitReleaseUnitTest' in unit test step in 'release' workflow)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -236,6 +236,8 @@ dependencies {
     testImplementation "org.robolectric:robolectric:${Versions.robolectric}"
     testImplementation "org.mockito:mockito-core:${Versions.mockito}"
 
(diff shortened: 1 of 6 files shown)


cnevinc:
Max said Mockito now supports static method. Maybe i'll try that

cnevinc:
We can mock final now. Just still can work with static method

PR #1955, opened 2018-05-18
Use the value from Firebase, if disabled will return the default value

cnevinc:
I found a bug that getShareDialogLaunchTimeThreshold will not use the value from Firebase Remote Config. This PR fixed that.


chianger:
[uplift] flag to go into 2.1

PR #1957, opened 2018-05-21
Browser fragment lifecycle

pharspher:


app/src/main/java/org/mozilla/focus/activity/ScreenNavigator.java
```diff
@@ -54,7 +54,6 @@ public void raiseBrowserScreen(boolean animate) {
         logMethod();
 
         this.browserMediator.raiseBrowserScreen(animate);
-        this.browserMediator.setInForeground(true);
```
pharspher: Redundant since back stack changed listener will handle this

app/src/main/java/org/mozilla/focus/activity/ScreenNavigator.java
```diff
@@ -89,7 +88,6 @@ public void addHomeScreen(boolean animate) {
 
         this.mainMediator.clearAllFragment(false);
         this.mainMediator.showHomeScreen(animate, true);
-        this.browserMediator.setInForeground(false);
```
pharspher: Redundant since back stack changed listener will handle this

app/src/main/java/org/mozilla/focus/activity/ScreenNavigator.java
```diff
@@ -100,6 +98,7 @@ public void popToHomeScreen(boolean animate) {
 
         this.mainMediator.clearAllFragment(false);
         this.mainMediator.showHomeScreen(animate, false);
+        this.browserMediator.setInForeground(false);
```
pharspher: popToHomeScreen() runs transaction without adding it to back stack, so we have to explicitly set browser fragment to background state here.

mTwTm:
Just noticed that what I previously state only happens when we have a clean install.

pharspher:
It will happen after FirstRunFragment was shown and dismissed.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -743,6 +743,11 @@ public void onBackPressed() {
             return;
         }
 
+        if (mainMediator.shouldFinish()) {
```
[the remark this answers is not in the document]
pharspher: I'll fix this in another PR.

app/src/main/java/org/mozilla/focus/tabs/tabtray/TabTrayFragment.java
```diff
@@ -251,6 +250,9 @@ public void closeTabTray() {
 
     @Override
     public void navigateToHome() {
+        if (isOpenedByHome()) {
```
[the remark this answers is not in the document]
pharspher: As discussed, steps to improve this would be
1. Make geo dialog a view instead of AlertDialog
2. Remove TabTrayFragment.FRAGMENT_TAG from MainMediator#FRAGMENTS_SEQUENCE
3. Move this protection stuff from TabTrayFragment to ScreenNavigator#addHomeScreen()

PR #1970, opened 2018-05-22
Use Kotlin in the real world

cnevinc:
ktx only works for API 28+.
So we just use pure Koltin.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -211,6 +211,8 @@ dependencies {
 
     implementation project(':firebase')
 
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${Versions.kotlin}"
+
     implementation "com.android.support:support-v4:${Versions.support}"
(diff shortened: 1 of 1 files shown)


cnevinc:
Confirmed KTX is only for 28+

cnevinc:
ktx only works for API 28+.
So we just use pure Koltin.

This PR still lacks:
1. Kotlin development guideline
2. Check style & find bug should still work
3. lint check for Kotlin
4. A file where extension functions should be placed. We don't want them to be anywhere and hard to be traced. Suggestion:
 a. If it's a custom class ( the class we wrote), it should be in the same file of the receiver class.
 b. If it extends a build in class (the class came from AOSP, libraries, Java Lang etc ) should under ... maybe a file name Extensions.kt under package root.

cnevinc:
Thanks, @mxlius
I'll evaluate and maybe do it in another patch.

PR #2023, opened 2018-05-29
Avoid clearing tabs persistence data without inserting the new tabs

benjamin-cheng:


benjamin-cheng:
Use `try` clause is to make sure the transaction is closed properly no matter if there is any exception happened. Similar to the try-with-resource, however, transaction is not Closeable. Room also implemented the database transaction in this pattern.

screen shot 2018-06-01 at 9 57 21 am

However, I found a better way to implement database transaction with Room's annotation. Please help to review, thank you!

Issue #2042, opened 2018-06-02
add ktlint


cnevinc:
before it's too late!

cnevinc:
Android Studio has official support. We'll use it
https://android-developers.googleblog.com/2018/03/android-studio-3-1.html

PR #2043, opened 2018-06-02
Issues #2042: Add ktlint

cnevinc:
Tested currently ./gradlew ktlint will pass.
Should we add this before it's too late?


cnevinc:
Android Studio has official support. We'll use it
https://android-developers.googleblog.com/2018/03/android-studio-3-1.html

Issue #1087, opened 2017-10-10
Consider adding ABPindo into the blocklist


brampitoyo:
ABPindo (https://github.com/ABPindo/indonesianadblockrules) an actively maintained blocklist for Indonesian websites. Adopting this list might help solve some of our site compatibility issues and make our Turbo Mode even better – but keep in mind that we may need to clear things with Legal before we do this.

weslyhuang:
next step on ENG to implement.

weslyhuang:
only partially support. agreed can go first with what we can do then see.

chianger:
close since #2051 was merged

PR #2051, opened 2018-06-04
Import ABPIndo list[adservers, prank, scam] to TurboMode

mxlius:
Closes, #1087


app/src/main/res/values/preference_keys.xml
```diff
@@ -9,6 +9,7 @@
     <string name="pref_key_privacy_block_analytics" translatable="false"><xliff:g id="preference_key">pref_privacy_block_analytics</xliff:g></string>
     <string name="pref_key_privacy_block_social" translatable="false"><xliff:g id="preference_key">pref_privacy_block_social</xliff:g></string>
     <string name="pref_key_privacy_block_other" translatable="false"><xliff:g id="preference_key">pref_privacy_block_other</xliff:g></string>
+    <string name="pref_key_privacy_block_abpindo" translatable="false"><xliff:g id="preference_key">pref_privacy_block_abpindo</xliff:g></string>
```
mTwTm: This may exceed the length limit that is enforced in a patch that @cnevinc just landed. Could be a problem after rebasing.

app/src/testWebkit/java/org/mozilla/focus/webkit/matcher/DisconnectTest.java
```diff
@@ -53,7 +53,7 @@ public void cleanup() {
     // http://robolectric.org/getting-started/#note-for-linux-and-mac-users
     @Test
     public void matches() throws Exception {
-        final UrlMatcher matcher = UrlMatcher.loadMatcher(RuntimeEnvironment.application, R.raw.blocklist, new int[] { R.raw.google_mapping }, R.raw.entitylist);
+        final UrlMatcher matcher = UrlMatcher.loadMatcher(RuntimeEnvironment.application, R.raw.blocklist, new int[] { R.raw.google_mapping }, R.raw.entitylist, R.raw.abpindo_adservers);
```
mTwTm: NIT: It's Disconnect and APDIndo test now. Also, maybe we can add some site to test this here too but I have no idea how we could start.

shavar-prod-lists/abpindo_adservers.json
```diff
@@ -0,0 +1,993 @@
+{
+    "categories": {
```
mTwTm: Extra unused file? Also, if we're using some library to process the json, consider adding them into tools/

app/src/testWebkit/java/org/mozilla/focus/webkit/matcher/DisconnectTest.java
mxlius: I agree, maybe wrap three lists(block, white, entity) into one source. Then pass-in a list with two sources(Disconnect, ABPIndo). Probably do this in next patch.

app/src/main/res/values/preference_keys.xml
mxlius: What's the length limitation? 40char? I guess analytics will exceed the boundary first?
Maybe we should write a custom lint to ensure the char length?

shavar-prod-lists/abpindo_adservers.json
mxlius: This is the source of the blocking list json file, gradle build copy this file into app/src/webkit/res/raw/. This is a hard-coded, handcraft, manual parsed JSON from the ABP format rules(https://adblockplus.org/filter-cheatsheet). I think we do need to write a parser if we want to keep the list updated or we need to add extra ABP region.

PR #2052, opened 2018-06-05
Upgrade Adjust and its companion install referrer SDK preparing for the invite referrer feature

benjamin-cheng:
Follow the Adjust SDK document => https://github.com/adjust/android_sdk#is-my-broadcast-receiver-capturing-the-install-referrer

The following screenshot is the log result while testing the install referrer broadcast
screen shot 2018-06-05 at 9 48 06 am


app/build.gradle
```diff
@@ -232,6 +232,7 @@ dependencies {
     implementation "org.mozilla.components:telemetry:${Versions.telemetry}"
 
     implementation "com.adjust.sdk:adjust-android:${Versions.adjust}"
+    implementation "com.android.installreferrer:installreferrer:${Versions.installreferrer}"
```
benjamin-cheng: Google Play Install Referrer Library
https://developer.android.com/google/play/installreferrer/library

cnevinc: Do we need data review for this? Did this use location services?
This may not be an issue for now, but we may need to keep a note on this..

PR #2064, opened 2018-06-15
Nevin/uitest fix

cnevinc:
Refine tests


app/src/androidTest/java/org/mozilla/focus/activity/BrowsingIntentTest.java
```diff
@@ -153,7 +140,8 @@ private void sendBrowsingIntent() {
         Intent intent = new Intent();
         intent.setAction(Intent.ACTION_VIEW);
         intent.setData(Uri.parse(TARGET_URL_SITE_1));
-        intent.setPackage(InstrumentationRegistry.getInstrumentation().getTargetContext().getPackageName());
-        activityTestRule.launchActivity(intent);
+        final Context targetContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        intent.setPackage(targetContext.getPackageName());
+        targetContext.startActivity(intent);
```
benjamin-cheng: Just curious about why use `targetContext.startActivity(intent)` instead of `activityTestRule.launchActivity(intent)`?

app/src/main/java/org/mozilla/focus/screenshot/CaptureRunnable.java
```diff
@@ -98,7 +93,13 @@ private void promptScreenshotResult(int snackbarTitleId) {
             return;
         }
         Toast.makeText(context, snackbarTitleId, Toast.LENGTH_SHORT).show();
-        completed = true;
+        if (refBrowserFragment != null &&
```
benjamin-cheng: It looks `refBrowserFragment.get() != null` check is duplicated. How about simplifying to 

if (refBrowserFragment != null) {
            final BrowserFragment browserFragment = refBrowserFragment.get();
            if (browserFragment != null && browserFragment.getCaptureStateListener() != null) {
                browserFragment.getCaptureStateListener().onPromptScreenshotResult();
            }
       }

cnevinc: fixed

app/src/androidTest/java/org/mozilla/focus/activity/BrowsingIntentTest.java
cnevinc: cause activityTestRule has already called launchActivity at this time. (see receivedBrowsingIntent_tabIsCreated). Here we only want to simulate third-party app sending browsing url intent to rocket.

PR #2069, opened 2018-06-20
Support color theme function, closes 1327

mxlius:


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -105,31 +107,25 @@ public View onCreateView(@NonNull LayoutInflater inflater,
...
-            }
+        home_container.setOnSwipeListener(new GestureListenerAdapter());
 
-            @Override
-            public void onSwipeDown() {
-                fakeInput.performClick();
-            }
-        });
+        if (ThemeManager.shouldShowOnboarding(view.getContext())) {
+            LayoutInflater.from(view.getContext()).inflate(R.layout.fragment_homescreen_themetoy, home_container);
```
mTwTm: Maybe themeOnboardingLayer = LayoutInflater.from(view.getContext()).inflate(R.layout.fragment_homescreen_themetoy, home_container); ?

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -162,9 +158,16 @@ public void onStop() {
         }
     }
 
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+        doWithActivity(getActivity(), themeManager -> themeManager.subscribeThemeChange(homeScreenBackground));
```
mTwTm: I tend to think that onStart() / onStop() are a better place for these. Or maybe onCreateView() / onDestroyView().

app/src/main/java/org/mozilla/focus/utils/OnSwipeListener.java
```diff
@@ -14,6 +14,12 @@
 
     void onSwipeDown();
 
+    void onLongPress();
+
+    boolean onDoubleTap();
+
+    boolean onSingleTapConfirmed();
+
     class OnSwipeListenerAdapter implements OnSwipeListener {
```
mTwTm: Consider using `default` modifier in `OnSwipeListener` methods so we don't need this Wrapper.

app/src/main/java/org/mozilla/rocket/theme/ThemeManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+import org.mozilla.focus.R;
+
+import java.util.HashSet;
+
+public class ThemeManager {
+
+    private static final int ONBOARDING_VERSION = 1;
+
+    private static final String PREF_KEY_STRING_CURRENT_THEME = "pref_key_string_current_theme";
```
mTwTm: I forgot what was the conclusion. Should we move these to the xml?

app/src/main/java/org/mozilla/rocket/theme/ThemeManager.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String PREF_KEY_INT_ONBOARDING_VERSION = "pref_key_int_onboarding_version";
+
+    public interface Themeable {
+        void onThemeChanged();
```
mTwTm: onThemeChanged(int themeId) could be a better option IMO.

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
mxlius: The return value of the ```inflate``` is the second param, which is the ```home_container``` if it is not ```null```. What I need is the ```themeOnboardingLayer```. When I pass ```home_container```, the inflated view will be add to the root also. so I can find the view with the id ```themeOnboardingLayer``` later.

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
mxlius: I think onActivityResult is the closet hook after onCreateView, and the subscription needs to happen before Fragment becomes visible to make sure all notify will be handled. It could possibly better if we do this in onCreateView, but I tend to make onCreateView to only View creation things.

app/src/main/java/org/mozilla/rocket/theme/ThemeManager.java
mxlius: I think this preference is more like a status config, not user modifiable settings. It's not used as a key of PreferenceItem. We probably should add a button in settings to allow user reset all the onboarding to learn what feature/on-boarading has been missing. Then, we can inject this key to that onboarding manager. Would that be better? @mTwTm

app/src/main/java/org/mozilla/rocket/theme/ThemeManager.java
mxlius: I'm okay if we add the ```themeId``` as a param. Just curious that if we should let the Themeable know what is the changed theme? Or we only notify and they just set the value from latest themed resources, no matter what the ```themeId``` is?

app/src/main/res/values-v23/theme_toy.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="ThemeToy.Light">
```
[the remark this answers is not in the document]
mxlius: This style does nothing for now but keeps a hint that these color will eventually need a Light/Dark design in the future for API 23 & above, because the statusbar style does not look good when we change background theme color to the dark one.

app/src/main/java/org/mozilla/rocket/theme/ThemeManager.java
mTwTm: Per what I've heard from @cnevinc he seems to believe these prefs should stay there too. At least that's what he told me when I was implementing unread screenshots. I think showing users the old features is a good idea but controlling it with PREF_KEY_STRING_CURRENT_THEME sounds a bit cumbersome to me. We'll need to handle more pairs of from-versions and to-versions. Maybe we should simply add them into settings so users can read anything anytime? Well, this may not work well with home onboarding and screenshot onboarding which are not shown in Cardviews, though.

mTwTm: My preference on this topic is I don't think it really matters where that key resides. It's more like a coding style to me so we can change at any time and I'm just wondering if we have that rule yet.

Issue #1104, opened 2017-10-13
Bookmark support


joechengla:

weslyhuang:
UX spec ready, pending on Visual one's (measures, ...)

weslyhuang:
Visual work is complete and the only left work is the color selection, per today's standup update. @Jacklin-ux please feel free to remove the "UX&VD Wanted" label once the selection is done : )

Jacklin-ux:
UX spec: https://drive.google.com/file/d/1R3QzuAM80-IaXLRMopYTupee2hjHQRWD/view?usp=sharing

mTwTm:
Should be closed with current integrations.

PR #2072, opened 2018-06-21
Issues/1104/ui part1

mTwTm:
Implement most of the UIs that has new strings.


mTwTm:
Part of #1104

app/src/main/java/org/mozilla/focus/activity/EditBookmarkActivity.kt
```diff
@@ -27,12 +27,41 @@ class EditBookmarkActivity : AppCompatActivity() {
     private val bookmark: TempInMemoryBookmarkRepository.Bookmark by lazy { TempInMemoryBookmarkRepository.getInstance().get(itemId) }
     private val name: EditText by lazy { findViewById<EditText>(R.id.bookmark_name) }
```
pharspher: I feel this should be renamed to something that can self-explain it's a UI widget, so one can easily tell it's not something parallel to another member variable "originalName", which is of type String.

app/src/main/java/org/mozilla/focus/activity/EditBookmarkActivity.kt
```diff
@@ -27,12 +27,41 @@ class EditBookmarkActivity : AppCompatActivity() {
...
     private val location: EditText by lazy { findViewById<EditText>(R.id.bookmark_location) }
-    private val originalName: String by lazy { name.text.toString() }
-    private val originalLocation: String by lazy { location.text.toString() }
+    private val originalName: String by lazy { bookmark.name }
+    private val originalLocation: String by lazy { bookmark.address }
     private lateinit var save: MenuItem
```
pharspher: Similar to the previous comment, the name "save" is quite ambiguous and it's hard to tell what one can do to this variable simply by reading its name.

PR #2108, opened 2018-06-29
Issues/1104/bug/minor issues

mTwTm:


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -191,4 +191,21 @@ public static String stripCommonSubdomains(@Nullable String host) {
...
+    public static String stripHttp(@Nullable String host) {
+        if (host == null) {
+            return null;
+        }
+        // In contrast to desktop, we also strip mobile subdomains,
+        // since its unlikely users are intentionally typing them
+        int start = 0;
+
+        if (host.startsWith("http://")) {
+            start = 7;
```
benjamin-cheng: It would be nice to define `http://` and `https://` as string constant and use `String.lenght()` to avoid the magic number.

PR #2155, opened 2018-07-09
Nevin/feature/incognito

### LINKED ISSUE #1909: Private browsing mode
Accounting to "Report for Firefox Rocket" user feedback, many users would like to have private browsing mode in our Rocket.

Pending bugs
----------------
#2244 Add Telemetry

cnevinc:
For #1909
Will land this first behind a pref (only visible for non-release and non-beta build)


app/src/main/res/xml/settings.xml
```diff
@@ -65,4 +66,9 @@
         android:title="@string/menu_about"
         android:key="@string/pref_key_about" />
```
walkingice: I feel like we could add a preference category here for *experimental*, and hide the category if it is under **release** or **beta** build

app/src/main/java/org/mozilla/rocket/incognito/IncognitoCore.kt
```diff
@@ -0,0 +1,16 @@
+package org.mozilla.rocket.incognito
+
+import android.content.Context
+import android.preference.PreferenceManager
+
+class IncognitoCore {
+    companion object {
+        const val PREF_KEY_INCOGNITO_ENABLED = "pref_key_incognito_enabled"
+
+        private const val PREF_KEY_INCOGNITO_ENABLED_DEFAULT = false
```
walkingice: should we set default value in xml file?

https://developer.android.com/guide/topics/ui/settings#Defaults

app/src/main/res/xml/settings.xml
cnevinc: I'd do this later in another commit/PR

PR #2167, opened 2018-07-12
Privately/refactor tabs session

walkingice:
Move tabs-related implementation to new module


components/browser/tabs/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">tabs</string>
+</resources>
```
pocmo: Looks like this is unused/unnecessary?

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -91,6 +85,13 @@
 import java.util.List;
 import java.util.Locale;
 
+import mozilla.components.tabs.Tab;
+import mozilla.components.tabs.TabModel;
+import mozilla.components.tabs.TabView;
+import mozilla.components.tabs.TabViewProvider;
+import mozilla.components.tabs.TabsSession;
+import mozilla.components.tabs.TabsSessionProvider;
```
pocmo: I recommend to not overload the `mozilla.components.*` package and leave this namespace to code that is actually part of a component. If this should end up in a component eventually then maybe use something like `incubator.mozilla.components.*` until it's moved into a component.

components/browser/tabs/src/test/java/mozilla/components/tabs/TabsSessionTest.java
```diff
@@ -187,7 +190,8 @@ public void testAddTab4() {
         Assert.assertEquals(session.getFocusTab().getId(), tabId1);
     }
 
-    @Test
+    // comment out this test case, I have no idea why it does not work for now
+    //@Test
```
pocmo: There's also `@Ignore` :)

components/browser/tabs/src/main/res/values/strings.xml
walkingice: Ok, removed

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
walkingice: I relocate the implementations to `org.mozilla.rocket.tabs`

PR #2187, opened 2018-07-18
Privately/add empty browser fragment

walkingice:
This PR added basic UI infrastructure to Private mode.

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
+++ b/app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
@@ -9,11 +9,14 @@ import android.content.Intent
 import android.os.Bundle
 import android.support.v4.app.Fragment
 import android.view.View
+import androidx.navigation.Navigation
 import org.mozilla.focus.R
 import org.mozilla.focus.activity.MainActivity
 import org.mozilla.focus.locale.LocaleAwareAppCompatActivity
+import org.mozilla.focus.urlinput.UrlInputFragment
 import org.mozilla.focus.widget.FragmentListener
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -25,34 +28,37 @@ class PrivateModeActivity : LocaleAwareAppCompatActivity(), FragmentListener {
...
-        val fragment = this.createHomeFragment()
-        this.supportFragmentManager.beginTransaction()
-                .add(R.id.container, fragment, PrivateHomeFragment.FRAGMENT_TAG)
-                .commit()
-    }
-
-    private fun createHomeFragment(): PrivateHomeFragment {
-        return PrivateHomeFragment.create()
+    override fun onBackPressed() {
+        val cnt = supportFragmentManager.backStackEntryCount
```
cnevinc: maybe add a comment about "Navigation can't work with backstack. if we have some manually added backstack, we must handle it before we forward the back key event to Navigation

app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -0,0 +1,81 @@
...
+    fun loadUrl(url: String?) {
+        url?.let {
+            if (it.isNotBlank()) {
+                displayUrlView.text = url
+            }
+        }
+    }
+
+    private fun registerData() {
+        val shared = ViewModelProviders.of(activity!!).get(SharedViewModel::class.java)
```
cnevinc: you probably want to remove the !!.

PR #2191, opened 2018-07-18
L10n export 2018 07 18

mTwTm:


mTwTm:
It's failing because there's a problematic string in a7e9088. This pr is just for the record. If l10n pr is merged we'll no longer need this.

mTwTm:
I redid it and the problematic string is fixed. Merging.

PR #2197, opened 2018-07-19
Privately/init session and handle back key

walkingice:


app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -61,6 +72,15 @@ class PrivateModeActivity : LocaleAwareAppCompatActivity(), FragmentListener {
         }
     }
 
+    override fun getTabsSession(): TabsSession {
+        if (session == null) {
+            session = TabsSession(PrivateTabViewProvider(this))
+        }
+
+        // we just created it, it definitely not null
```
pharspher: it seems this method can be removed by making seesion a lazy property?

walkingice: if we assign `lateinit ` to it, Kotlin will complain on `if (session == null)`

pharspher: What I''m talking about is lazy delegate like
val session: TabSessions by lazy { TabSessions(...) }

PR #2207, opened 2018-07-21
Issue 2189 tab mode validation

walkingice:
Rewrite `TabModel.isValid`


components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/Tab.java
```diff
@@ -51,7 +51,9 @@ public TabModel getSaveModel() {
 
         if (tabView != null) {
             tabModel.setTitle(tabView.getTitle());
-            tabModel.setUrl(tabView.getUrl());
+            if (!TextUtils.isEmpty(tabView.getUrl())) {
```
mTwTm: Maybe we don't need to ignore this.

PR #2208, opened 2018-07-21
Nevin/feature/privately easy

cnevinc:


walkingice:
I still feel like `privately.SharedViewModel` is easier than `privately.PrivateModeViewModel`. but it's on you.

PR #2210, opened 2018-07-22
Privately/context menu

walkingice:


cnevinc:
Just wondering, if we always use context.getApplicationContext and cast then check if we are in PrivateMode (using Context Wrapper to wrap application context) will that be an easier way to check if we are in PrivateMdoe or not?

Issue #2243, opened 2018-07-24
Cannot fullscreen YouTube video in private browsing mode


joechengla:
screen will seem freeze

walkingice:
close this issue since the PR has been merged

cynthiatang:
Verified: Fixed

Nexus5 
Android 5.1.0
Rocket 3.0.0(5988).nightly
WebView 67.0.3396.87

PR #2278, opened 2018-07-26
privately: add container for fullscreen video

walkingice:
Port the fullscreen related implementation from BrowserFragment.java.
Currently we don't handle screen rotation since it will effect Fragment
re-creation.

to solve #2243


app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -222,6 +227,9 @@ class BrowserFragment : LocaleAwareFragment(),
     }
 
     class BrowserTabsChromeListener(val fragment: BrowserFragment) : DefaultTabsChromeListener() {
```
cnevinc: What's the difference if we use inner class here?

walkingice: There are not many technical difference, I just want to explicit specify what parts belong to fragment. I am OK if anyone want to change it to inner class.

PR #2285, opened 2018-07-27
Add fastlane screengrab support

josephyu1110:


app/build.gradle
```diff
@@ -245,6 +245,7 @@ dependencies {
     })
     androidTestImplementation "com.android.support.test.espresso:espresso-web:${Versions.espresso}"
     androidTestImplementation "com.android.support.test.espresso:espresso-intents:${Versions.espresso}"
+    androidTestImplementation 'tools.fastlane:screengrab:1.1.0'
```
cnevinc: use ${Versions.xxxx}" instead of a number

app/src/androidTest/java/org/mozilla/focus/helper/BeforeTestTask.java
```diff
@@ -51,6 +52,7 @@ public void execute() {
         }
 
         Inject.getTabsDatabase(null).tabDao().deleteAllTabs();
+        BookmarksDatabase.getInstance(context).bookmarkDao().deleteAllBookmarks();
```
cnevinc: Why is this needed? shouldn't all bookmarks be empty when the app first launched?

app/src/androidTest/java/org/mozilla/focus/screengrab/AaHomeScreenshot.java
```diff
@@ -0,0 +1,150 @@
...
+        onView(withId(R.id.menu_bookmark)).perform(click());
+        onView(withText(R.string.bookmarks_empty_view_msg)).check(matches(isDisplayed()));
+        Screengrab.screenshot(ScreenshotNamingUtils.HOME_MENU_NO_BOOKMARKS);
+        SystemClock.sleep(MockUIUtils.TAKE_SCREENSHOT_DELAY);
+
+        // Click downloads
+        onView(withId(R.id.downloads)).perform(click());
+        onView(withText(R.string.no_downloads)).check(matches(isDisplayed()));
+        Screengrab.screenshot(ScreenshotNamingUtils.HOME_MENU_NO_DOWNLOAD);
+        SystemClock.sleep(MockUIUtils.TAKE_SCREENSHOT_DELAY);
```
cnevinc: I don't like the delay here. Should be a better way for the delay.
Talked to Joesph, This is because Screengrab won't wait for Idling Resources.
Maybe we should use https://developer.android.com/reference/android/support/test/runner/screenshot/Screenshot 
And switch locales using the pref in our settings.

cnevinc:
maybe add below code and some comment to describe it
```
<uses-permission android:name="android.permission.READ_PRIVILEGED_PHONE_STATE"
    tools:ignore="ProtectedPermissions" />
```

cnevinc:
I'm not sure how to set testInstrumentationRunnerArguments in Bitrise so I fill [this bug](https://github.com/bitrise-steplib/steps-virtual-device-testing-for-android/issues/35)

Currently I set it in our CustomTestRunner

cnevinc:
Switching between test runners now seems to work. 
Now I tried to put the generated screenshots in the download files section.

cnevinc:
Now testing on CI.
Since we need to use "fastlane screenshot" to execute the task, we can't use "Virtual Device Test" task in Bitrise. The work around is to manually create avd , and start it , then run the command.

The problem till now is Bitrise only have API 24 for x86 avd. API 24 is only for arm avd. 

The test keeps falling on Bitrise so we want to exclude the API issue first.

I'll try using the arm 24 avd. Then report back.

If it still didn't work, we'll need screengrab work for API 26.

It should be running no matter what API level it is. any way.

PR #2295, opened 2018-07-31
Nevin/domain

cnevinc:
Adding Android-components-domains.
@pocmo , please advise :)


app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.java
```diff
@@ -94,8 +94,7 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         clearView = view.findViewById(R.id.clear);
         clearView.setOnClickListener(this);
 
-        urlAutoCompleteFilter = new UrlAutoCompleteFilter();
-        urlAutoCompleteFilter.loadDomainsInBackground(getContext().getApplicationContext());
+        urlAutoCompleteFilter = new UrlAutoCompleteFilter(getContext());
```
walkingice: the original author specify to use Application-Context here, I think we should follow the enlightenment come from the one.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -20,7 +20,7 @@ object Versions {
     const val findbugs = "3.0.1"
     const val lottie = "2.5.5"
     const val leakcanary = "1.5.4"
-    const val telemetry = "0.5.1"
+    const val android_components = "0.10"
```
pocmo: Is there a particular reason why you are using 0.10 and not the latest release (0.16.1)?

app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
```diff
@@ -6,124 +6,27 @@
...
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Locale;
-import java.util.Set;
+import mozilla.components.browser.domains.DomainAutoCompleteProvider;
 
 public class UrlAutoCompleteFilter implements InlineAutocompleteEditText.OnFilterListener {
```
pocmo: You may be interested in using `ui-autocomplete` eventually. This will replace `InlineAutocompleteEditText`. For one the version component will replace `OnFilterListener` with a simple function (yay!) and more importantly we landed some bug fixes! :)

buildSrc/src/main/java/Dependencies.kt
cnevinc: Will do!

app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
cnevinc: Will do it in another PR. Thanks!

PR #2298, opened 2018-07-31
Privately/purify tabs

walkingice:


app/src/main/java/org/mozilla/rocket/privately/PrivateTabViewProvider.kt
```diff
@@ -12,6 +13,11 @@ class PrivateTabViewProvider(private val host: Activity) : TabViewProvider {
         return WebViewProvider.create(host, null, WebViewSettingsHook) as TabView
     }
 
+    override fun purify() {
+        create().cleanup()
+        CookieManager.getInstance().removeAllCookies(null)
```
mxlius: Looks like WebkitView.cleanup already includes CookieManager.removeAllCookies. Do we still need to do it here?

walkingice: I was thinking to do that explicitly, but remove that also make sense to me.

app/src/main/java/org/mozilla/rocket/privately/PrivateTabViewProvider.kt
```diff
@@ -12,6 +12,10 @@ class PrivateTabViewProvider(private val host: Activity) : TabViewProvider {
         return WebViewProvider.create(host, null, WebViewSettingsHook) as TabView
     }
 
+    override fun purify() {
+        create().cleanup()
```
cnevinc: Does this create a new Tabview every time a client call purify()?

walkingice: yes, why?

walkingice: Nevin was worrying about performance waste. We will refactor `TabView.cleanup()` soon it should be OK for now.

PR #2301, opened 2018-08-01
Nevin/autocomplete

cnevinc:
ready for review
We use a-c's autocomplete version.


cnevinc:
Autocompletes use coroutine to load assets files. So the increase in APK size is unavoidable.
[see the code](https://github.com/cnevinc/android-components/blob/master/components/browser/domains/src/main/java/mozilla/components/browser/domains/DomainAutoCompleteProvider.kt#L120)
According to the log, it's 3096343 > 2936012.8

cnevinc:
I'll add 160KB to the APK size for coroutine in a-c autocomplete

PR #2310, opened 2018-08-03
[Android Components Incubator] TabCounter 

cnevinc:
Ready for review,
After landed, will test it on Release channel and request to merge into Android-components

@pocmo Please help advise. Thanks!


cnevinc:
APK size 爆惹... QQ

mTwTm:
I took a look at the components size difference between 10 and 16, no big difference. I think the oversized apk is because you did not minifyEnabled false in ui-tabcounter/build.gradle?

settings.gradle
```diff
@@ -1,4 +1,4 @@
-include ':feature-tabs'
+include ':feature-tabs', ':ui-tabcounter'
```
walkingice: separate into 2 lines?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -82,6 +81,9 @@
 import java.lang.ref.WeakReference;
 import java.util.WeakHashMap;
 
+import incubator.org.mozilla.components.ui.tabcounter.TabCounter;
+
+
```
walkingice: redundant new line

settings.gradle
walkingice: separate to this?
```
include ':feature-tabs'
include ':ui-tabcounter'
```

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
cnevinc: fixed

pocmo:
> Don't use utils from a-c to make Rocket APK smaller

@cnevinc That's interesting. I would have expected proguard to remove everything except the single helper method you use. Do you remember how big the difference is?

cnevinc:
Firstly I have the same thought as you. But I look at the APK difference I found kotlinx/coroutine and more kotlin classes I didn't use. 
I'm still investigating 
image

PR #2333, opened 2018-08-06
Replace topsties http links with https

### LINKED ISSUE #2316: Replace topsties http links with https
https should be safer to protect what we promote.

mTwTm:
Fixes #2316


mTwTm:
Since tribunnews and gmanetwork has mixed contents. Probably not a good idea to merge this.

mTwTm:
Bitrot. Also, there are some webcompat showstoppers as earlier commented.

PR #2337, opened 2018-08-06
Setup firebase banner cache

### LINKED ISSUE #2324: The second load of Banner should load instantly.

mTwTm:
Fixes #2324

### DIFF
--- a/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
+++ b/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
@@ -48,6 +48,7 @@ import org.mozilla.focus.provider.QueryHandler;
 import org.mozilla.focus.tabs.TabCounter;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.AppConfigWrapper;
+import org.mozilla.focus.utils.FileUtils;
 import org.mozilla.focus.utils.FirebaseHelper;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -201,29 +201,67 @@ protected void onPostExecute(String line) {
...
-                BannerAdapter bannerAdapter = new BannerAdapter(configArray, arg -> FragmentListener.notifyParent(this, FragmentListener.TYPE.OPEN_URL_IN_NEW_TAB, arg));
-                banner.setAdapter(bannerAdapter);
-                banner.setLayoutManager(new LinearLayoutManager(getContext(), RecyclerView.HORIZONTAL, false));
-                SnapHelper snapHelper = new PagerSnapHelper();
-                snapHelper.attachToRecyclerView(banner);
-                showBanner(true);
-            } catch (JSONException e) {
-                Logger.throwOrWarn(TAG, "Invalid Config: " + e.getMessage());
-            } finally {
+                writeToCache(context, configArray);
```
mxlius: It seems we are doing File IO on the main thread. Maybe we can file another PR to move this into LiveData pattern?

mTwTm: I think this is very lightweight I/O so should be okay. But I'll file a bug.

PR #2347, opened 2018-08-07
Issues/1798

### LINKED ISSUE #1798: Experiment: Asian Games in Jakarta content offering

mTwTm:
Fixes #1798


mTwTm:
@mxlius Can you take special care with ee0f914 given that I'm afraid there will be some leak / performance problem.

PR #2358, opened 2018-08-08
Autoplay banner slideshow

### LINKED ISSUE #2325: Banner should scroll automatically every 10 seconds.

mTwTm:
Fixes #2325


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -340,6 +364,8 @@ public void onPause() {
         if (context != null) {
             LocalBroadcastManager.getInstance(context).unregisterReceiver(this.receiver);
         }
+        timer.cancel();
```
mxlius: Do we resume the timer when activity onResume?

PR #2359, opened 2018-08-08
Safe check for future unupdated users: Don't inflate unknown typed banner pages

### LINKED ISSUE #2353: Do not fallback unknown banner type.

mTwTm:
Fixes #2353

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
+++ b/app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
@@ -8,15 +8,23 @@ import android.view.ViewGroup;
 
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.mozilla.rocket.util.Logger;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
 
 public class BannerAdapter extends RecyclerView.Adapter<BannerViewHolder> {
 
-    @IntDef({BasicViewHolder.VIEW_TYPE, SingleButtonViewHolder.VIEW_TYPE, FourSitesViewHolder.VIEW_TYPE})
+    private static final String LOG_TAG = "BannerAdapter";
+
+    @IntDef({UNKNOWN_VIEW_TYPE, BasicViewHolder.VIEW_TYPE, SingleButtonViewHolder.VIEW_TYPE, FourSitesViewHolder.VIEW_TYPE})
     @interface ViewType {}
 
     private Context context;
-    private BannerDAO[] DAOs;
+    private List<BannerDAO> DAOs;
     private OnClickListener onClickListener;
+    private static final int UNKNOWN_VIEW_TYPE = -1;
 
     @Override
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
```diff
@@ -29,27 +33,39 @@ public void onDetachedFromRecyclerView(@NonNull RecyclerView recyclerView) {
...
+        switch (type) {
             case SingleButtonViewHolder.VIEW_TYPE_NAME:
                 return SingleButtonViewHolder.VIEW_TYPE;
             case FourSitesViewHolder.VIEW_TYPE_NAME:
                 return FourSitesViewHolder.VIEW_TYPE;
             case BasicViewHolder.VIEW_TYPE_NAME:
-            default:
                 return BasicViewHolder.VIEW_TYPE;
+            default:
+                return UNKNOWN_VIEW_TYPE;
```
mxlius: Since we already filter unknown type from input parsing, should we also throw an exception on non-release build? to raise the problem?

app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
```diff
@@ -62,18 +78,19 @@ public BannerViewHolder onCreateViewHolder(@NonNull ViewGroup parent, @ViewType
             case FourSitesViewHolder.VIEW_TYPE:
                 return new FourSitesViewHolder(parent, onClickListener);
             case BasicViewHolder.VIEW_TYPE:
+            case UNKNOWN_VIEW_TYPE:
```
mxlius: Since we already filter unknown type from input parsing, should we also throw an exception on non-release build? to raise the problem?

app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
```diff
@@ -29,27 +33,39 @@ public void onDetachedFromRecyclerView(@NonNull RecyclerView recyclerView) {
...
             BannerDAO thisDAO = new BannerDAO();
             thisDAO.type = jsonObject.getString(BannerDAO.TYPE_KEY);
             thisDAO.values = jsonObject.getJSONArray(BannerDAO.VALUES_KEY);
-            this.DAOs[i] = thisDAO;
+            // We always write full item into cache, but discard a page if it is not know in this
+            // apk version. With this once the user upgrades, the cache is still usable.
+            // We still inflate an item with SingleButtonViewHolder if it for unknown reason managed
+            // to pass this check. (See: onCreateViewHolder)
+            if (getItemViewType(thisDAO.type) == UNKNOWN_VIEW_TYPE) {
+                continue;
```
mxlius: Should we throw an exception on nightly?

mTwTm:
I think adding in the parsing block would be suffice, others will only happen when someone is breaking the parsing and adding some local items.

PR #2379, opened 2018-08-09
Apply UI l10n when editing bookmark

mxlius:


mTwTm:
Still wondering why android:label="@string/title_activity_edit_bookmark" in AndroidManifest is not working any more 🤔

mTwTm:
Confirmed this is not a regression and probably label itself is not applying because it happens before Activity's own lifecycle.

PR #2389, opened 2018-08-10
 Use Firebase to pref on/off private browsing

cnevinc:
We want to disable this feature after we ship.


app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -52,6 +53,7 @@
     private static final String FIREBASE_APP_ID = "google_app_id";
     private static final String FIREBASE_API_KEY = "google_api_key";
     private static final String FIREBASE_PROJECT_ID = "project_id";
+    private static final boolean BOOLEAN_DEFAULT = false;
```
mTwTm: Rename: BOOLEAN_DEFAULT?

PR #2396, opened 2018-08-11
WIP refine Private Mode

cnevinc:
Please advise!


cnevinc:
We won't take this approach.

PR #2411, opened 2018-08-13
[Closes #2291] Custom preferences ping for ThemeColor & CaptureCount

### LINKED ISSUE #2291: Can't find color theme data in core ping table
Below are my test queries:
https://sql.telemetry.mozilla.org/queries/57420#149719
https://sql.telemetry.mozilla.org/queries/52865#140594

Also do we have anything else we think should be in core-ping but we can't find?
@mxlius @cnevinc

mxlius:


cnevinc:
CaptureCountMeasurement is not really a setting, should we put it here?

PR #2443, opened 2018-08-17
Fix TakeScreenshotTest break issue.

josephyu1110:
- Test is failed when Lottie is updated from 2.2.0 to 2.5.5. We can observe the break since 2.5.0+.
- During TakeScreenshotTest is running, screen capture animation won't be played when Lottie version is less than 2.5.0 but is played on 2.5.0+.
- Test failed only happens on Bitrise and cannot reproduce locally.
- If the animation is played, TakeScreenshotTest will be stuck in:
https://github.com/mozilla-tw/Rocket/blob/37e309e79a0fbf711fe9a48e8f7c1cfad9dcf1da/app/src/main/java/org/mozilla/focus/screenshot/ScreenshotCaptureTask.java#L70-L71

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/Inject.java
+++ b/app/src/androidTest/java/org/mozilla/focus/Inject.java
@@ -82,4 +82,7 @@ public class Inject {
         // Do nothing in testing
     }
 
+    public static boolean isUnderEspressoTest() {
+        return true;
+    }
 }
--- a/app/src/main/java/org/mozilla/focus/Inject.java
+++ b/app/src/main/java/org/mozilla/focus/Inject.java
@@ -71,4 +71,8 @@ public class Inject {
     public static void setActivityNewlyCreatedFlag() {
         sIsNewCreated = false;
     }
+
+    public static boolean isUnderEspressoTest() {
+        return false;
+    }
 }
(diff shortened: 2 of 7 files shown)


app/src/androidTest/java/org/mozilla/focus/Inject.java
```diff
@@ -82,4 +84,7 @@ public static void setActivityNewlyCreatedFlag() {
         // Do nothing in testing
     }
 
+    public static boolean isScreenCaptureAnimationEnabled() {
+        return sIsScreenCaptureAnimationEnabled;
```
cnevinc: you can just return false here.

app/src/main/java/org/mozilla/focus/Inject.java
```diff
@@ -71,4 +73,8 @@ public static final boolean getActivityNewlyCreatedFlag() {
     public static void setActivityNewlyCreatedFlag() {
         sIsNewCreated = false;
     }
+
+    public static boolean isScreenCaptureAnimationEnabled() {
+        return sIsScreenCaptureAnimationEnabled;
```
cnevinc: you can just return true here.

app/src/main/java/org/mozilla/focus/fragment/ScreenCaptureDialogFragment.java
```diff
@@ -57,6 +58,11 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         View view = inflater.inflate(R.layout.dialog_capture_screen, container, false);
         mLottieAnimationView = (LottieAnimationView) view.findViewById(R.id.animation_view);
+        if (!Inject.isScreenCaptureAnimationEnabled()) {
```
cnevinc: can you inject the playback condition around playAnimation() instead of setRepeatCount()? So the original code flow will be similar. 
I also suggest renaming the method isScreenCaptureAnimationEnabled to isUnderEspressoTest for better understanding why we want to disable the animation in EspressoTest.

PR #2453, opened 2018-08-20
Add category for screenshot. closes #1987

### LINKED ISSUE #1987: Telemetry: Understand screenshot better
Without knowing the site URL, can we categorize and annonymize the types of website screenshots were taken on? Is it mostly shopping? News?... Etc

cnevinc:
We load the category mapping list in Application class, and use it without
making sure it's there. It's something we need to fix in the future.

Ideally, every config should be ready before we really use it.
We didn't do it here because
1. The schedule doesn't allow us.
2. This feature may be removed in the future.

After confirmed with UX and product, we'll display empty string is the
mapping is not ready.


app/src/main/java/org/mozilla/focus/screenshot/ScreenshotManager.java
```diff
@@ -52,4 +69,47 @@ public void update(org.mozilla.focus.screenshot.model.Screenshot screenshot, Asy
...
+        } catch (IOException | JSONException e) {
+            e.printStackTrace();
+        }
+    }
+
+    public String getCategory(String url) {
+        try {
+            if (categories.size() == 0) {
+                return "";
+            }
```
walkingice: there might be threading issue here. And here there are two cases: if no categories we return `""` empty string, if any exception we return `DEFAULT_CATEGORY`. Leave a comment for the difference would be helpful.

app/src/main/java/org/mozilla/focus/screenshot/ScreenshotManager.java
```diff
@@ -52,4 +69,47 @@ public void update(org.mozilla.focus.screenshot.model.Screenshot screenshot, Asy
...
+        }
+        final JSONObject json;
+        try {
+            json = IOUtils.readAsset(context, "screenshots-mapping.json");
+
+            final Iterator<String> iterator = json.keys();
+            while (iterator.hasNext()) {
+                final String category = iterator.next();
+                final Object o = json.get(category);
+                if (o instanceof JSONArray) {
```
walkingice: maybe `JSONObject.optJSONArray(String name)` could be applied here

PR #2465, opened 2018-08-23
Uri based site

### LINKED ISSUE #2438: Support bootstrapping top site icons from remote resources.
Currently, we're bootstrapping top sites with local configurations. We'd like to load these from remote to:

1. be more flexible

and

2. shrink apk size.

mTwTm:
Update(18/08/31):

Fixes #2438 

--

This is a WIP pull request to get things reviewed early.
Basically, this is the preliminary work of #2438. The original implementation of the Site class and the backing database only takes Bitmaps as a parameter. With this implementation, it would be complex to load files from remote servers. Furthermore, in-database blobs are something we long been eager to optimize.

Also, I've taken the change to migrate to Room but we're only using the schema, the loading is still utilizing traditional cursors rather than Daos.

I've also moved a lot files around to make things clearer. Please also suggest better ways to organize this if you see something wrong or if you want to keep the histories of certain files and I failed to do so.

There are some implementation gaps that still needed to be filled:
1. If users have local (non-default) top sites, the icons of these will disappear.
-> I'm planning to also write these blobs to file during migration, 8 sites shouldn't take long.
2. This draft implementation is not handling favicons that are too blurry to display. That is, we had a minimum size threshold to decide if we want to fall back to generated character badges. This will be take cared in a future version of this pull request.
3. I think I'd also need to check heavy i/o things are happening in the background.


mTwTm:
Ready to be reviewed :)

app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
```diff
@@ -89,11 +101,24 @@ public void onBindViewHolder(RecyclerView.ViewHolder holder, final int position)
                 siteVH.rootView.setOnClickListener(this);
                 siteVH.textMain.setText(item.getTitle());
                 siteVH.textSecondary.setText(item.getUrl());
-                Bitmap bmpFav = item.getFavIcon();
-                if (bmpFav != null) {
-                    siteVH.imgFav.setImageBitmap(bmpFav);
+                String favIconUri = item.getFavIconUri();
+                if (favIconUri != null) {
+                    Glide.with(siteVH.imgFav).load(favIconUri).into(siteVH.imgFav);
+                    Glide.with(siteVH.imgFav.getContext())
```
benjamin-cheng: Why load `favIconUri` twice? duplicated code?

mTwTm:
Fixes #2438

components/utils/icon/src/main/java/org/mozilla/icon/FavIconUtils.java
```diff
@@ -171,4 +183,136 @@ private static String getRepresentativeSnippet(@NonNull String url) {
...
+            if (!file.createNewFile()) {
+               if (file.exists()) {
+                   return Uri.fromFile(file).toString();
+               } else {
+                   return null;
+               }
+            }
+            //Convert bitmap to byte array
+            ByteArrayOutputStream bos = new ByteArrayOutputStream();
+            bitmap.compress(Bitmap.CompressFormat.PNG, 0 /*ignored for PNG*/, bos);
```
benjamin-cheng: How about save it as JPEG with maybe 80% quality to save disk space?

app/src/main/java/org/mozilla/focus/utils/DimenUtils.java
```diff
@@ -0,0 +1,77 @@
...
+    private static float getDefaultFaviconTextSize(Resources resources) {
+        return resources.getDimension(R.dimen.favicon_initial_text_size);
+    }
+
+    private static int getDefaultFaviconBitmapSize(Resources resources) {
+        return resources.getDimensionPixelSize(R.dimen.favicon_target_size);
+    }
+
+    public static Bitmap getInitialBitmap(Resources resources, Bitmap source, char initial) {
+        return FavIconUtils.getInitialBitmap(source, initial, getDefaultFaviconTextSize(resources), getDefaultFaviconBitmapSize(resources));
```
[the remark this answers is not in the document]
mTwTm: The idea is: DimenUtils is for per-app configurations. FavIconUtils should be more general.

app/src/main/java/org/mozilla/focus/utils/TopSitesUtils.java
```diff
@@ -104,7 +89,7 @@ public static Bitmap getIconFromAssets(Context context, String fileName) {
                     site.setViewCount(json_site.getLong("viewCount"));
                     site.setLastViewTimestamp(json_site.getLong("lastViewTimestamp"));
                     String icon_name = json_site.getString("favicon");
-                    site.setFavIcon(TopSitesUtils.getIconFromAssets(context, icon_name));
+                    site.setFavIconUri("file:///android_asset/topsites/icon/" + icon_name);
```
mxlius: nit: How about replacing the magic string with constants?

components/utils/icon/build.gradle
```diff
@@ -1,4 +1,7 @@
 apply plugin: 'com.android.library'
+apply from: '../../../buildSrc/pmd.gradle'
+apply from: '../../../buildSrc/checkstyle.gradle'
+apply from: '../../../buildSrc/findbugs.gradle'
```
mxlius: How about replacing the relative path with rootProject path?
"$rootProject.projectDir/buildSrc/findbugs.gradle"

PR #2472, opened 2018-08-24
Fix NavigationTest and BrowsingIntentTest flaky issue.

josephyu1110:
Fix test break in build:
https://app.bitrise.io/build/78680edaa48a299a
- BrowsingIntentTest:
android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. 
- NavigationTest
The symptom is calling Espresso.pressBack() will go back to HomeFragment instead of previous web page while web view did have back stack. Looks like some transaction is not ready once we call pressBack immediately when the web page is loaded. To avoid this case, we add some actions(swipe up and swipe down)  to simulate browsing the web page once the page is loaded, then call pressBack().

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingIntentTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingIntentTest.java
@@ -140,6 +140,7 @@ public class BrowsingIntentTest {
         Intent intent = new Intent();
         intent.setAction(Intent.ACTION_VIEW);
         intent.setData(Uri.parse(TARGET_URL_SITE_1));
+        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
         final Context targetContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
         intent.setPackage(targetContext.getPackageName());
         targetContext.startActivity(intent);


josephyu1110:
Below are some builds include this fix, original NavigationTest and BrowsingIntentTest failed cases are gone.
https://app.bitrise.io/build/aae79c974685f262
https://app.bitrise.io/build/27c435e8483045b2
https://app.bitrise.io/build/994d5eb9ec74ff64

app/src/androidTest/java/org/mozilla/focus/activity/NavigationTest.java
```diff
@@ -99,6 +101,9 @@ public void browsingWebsiteBackAndForward_backAndFrowardToWebsite() {
                 .check(matches(withText(TARGET_URL_SITE_2)));
         IdlingRegistry.getInstance().unregister(loadingIdlingResource);
 
+        // play around the web content view
+        onView(withId(R.id.main_content)).perform(swipeUp(), swipeDown());
```
cnevinc: what we really want is to get the correct timing to perform "go back"
This works but not ideal and is misleading.
Maybe we can just add a comment and a TODO here.
Please help file a new issue to track this work.

josephyu1110:
NavagationTest intermittent issue is tacked in #2575

mTwTm:
The build failed because I'm making some changes. I'll re-trigger.

PR #2477, opened 2018-08-24
Launch system default app directly when tapping default browser notif…

mxlius:
…ication

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
@@ -11,6 +11,7 @@ import android.app.PendingIntent;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.v4.app.NotificationCompat;
 import android.support.v7.app.AlertDialog;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
```diff
@@ -174,9 +175,16 @@ public static void showRateAppNotification(Context context) {
...
 
         // Let NotificationActionBroadcastReceiver handle what to do
-        final Intent openDefaultBrowserSetting = IntentUtils.genDefaultBrowserSettingIntentForBroadcastReceiver(context);
-        final PendingIntent openRocketPending = PendingIntent.getBroadcast(context, REQUEST_DEFAULT_CLICK, openDefaultBrowserSetting,
-                PendingIntent.FLAG_ONE_SHOT);
+        final Intent openDefaultBrowserSetting;
+        final PendingIntent openRocketPending;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            openDefaultBrowserSetting = new Intent(android.provider.Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS);
+            openRocketPending = PendingIntent.getActivity(context, REQUEST_DEFAULT_CLICK, openDefaultBrowserSetting, PendingIntent.FLAG_ONE_SHOT);
```
cnevinc: If we open the setting directly, can we still monitor/send Telemetry to track the click action of that notification if API level >= 24?
see:
https://github.com/mozilla-tw/Rocket/blob/567ebbe3b04dea34bc28200e4845da9e8354c952/app/src/main/java/org/mozilla/focus/notification/NotificationActionBroadcastReceiver.java#L73

mxlius: Yes, you are right. Good catch, the current implementation is as expected.

PR #2484, opened 2018-08-27
Keep screen on when playing video in private browsing mode. Closes #…

josephyu1110:
…2445

- Also hide status bar and allow screen rotation in fullscreen video view


cnevinc:
@cynthiatang Please help verify

PR #2506, opened 2018-08-30
pref: add Development category to preference

walkingice:
This category contains some preference only for development purpose,
hence it is only visible under dev-build.


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -42,10 +42,10 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         addPreferencesFromResource(R.xml.settings);
-        if (AppConstants.isReleaseBuild() || AppConstants.isBetaBuild()) {
+        if (!AppConstants.isDevBuild()) {
```
cnevinc: We may want to enable it firebase(debug) build as well

PR #2522, opened 2018-08-31
Deeplinks -> Goto Android Set Default App page (Android 7.0+) through Firebase 

cnevinc:
To sums up, this PR helps
1.  Move all launching behaviour all together into LaunchIntentDispatcher. by
    a. LaunchIntentDispatcher will check for extras and send telemetry if needed.
    b. RocketLauncherActivity will only handle if it needs to continue/open MainAcitvity
    c. Save parsing extras from MainActivity
2. Add the ability to handle app links from push notifications.

Next step will be
1. Make Rocket [support intent](https://developer.chrome.com/multidevice/android/intents)

2. I've found some dupicated TelemetryEvent (MainActivity.sendBrowsingTelemetry())
   I'll try to remove them when I have time.


cnevinc:
dup with #2539

PR #2544, opened 2018-09-05
Fip/basic ui

walkingice:
Include Max's implementation, and add basic UI of Find-in-page. It is only visible in debug-build


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -437,7 +436,6 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         }
         if (searchBtn != null) {
             searchBtn.setOnClickListener(this);
-            searchBtn.setOnLongClickListener(this);
```
cnevinc: Can we use method reference can wrap this line behind a pref? for example
```
if (pref.isFipEnabled()){
    searchBtn.setOnLongClickListener(this:: onLongClick);
}
```
}

walkingice: We removed the Long-click-listener, so it would be fine here.

PR #2554, opened 2018-09-06
Max/exp/partner customization

mxlius:


app/src/main/java/org/mozilla/rocket/partner/PartnerActivator.java
```diff
@@ -0,0 +1,317 @@
...
+                urlConnection.setDoInput(true);
+                urlConnection.setDoOutput(true);
+
+                urlConnection.setRequestProperty("Content-Type", "application/json");
+
+
+                JSONObject root = new JSONObject();
+                JSONObject fields = new JSONObject();
+                root.put("fields", fields);
+                fields.put("device_id", new JSONObject("{\"stringValue\":\"" + PartnerUtil.getDeviceIdentifier(partnerActivator.context.getContentResolver()) + "\"}"));
```
mTwTm: I'm not sure why we're wrapping the child with "{\"stringValue\"...
If we really need to do so, consider writing a function for doing it.

app/src/main/java/org/mozilla/rocket/partner/PartnerActivator.java
```diff
@@ -0,0 +1,317 @@
...
+                } catch (Exception e) {
+                    PartnerUtil.log(e, "PingActivation post exception");
+                }
+                int responseCode = urlConnection.getResponseCode();
+                if (responseCode != HttpURLConnection.HTTP_OK) {
+                    //  TODO: Telemetry ping fail
+                    PartnerUtil.log("PingActivation server response not OK: " + responseCode);
+                }
+
+                partnerActivator.setStatus(Status.Done);
```
mTwTm: status is set to done even if there are some exceptions.

PR #2570, opened 2018-09-10
Disable LaunchIntentDispatcherTest to unblock release build

mxlius:
This test failed with callstack in FirebaseAnalytics which blocks release process. Disable it until we have found the solution.


mTwTm:
Filed #2572. I'm okay with this given the urgent nature.

PR #2579, opened 2018-09-12
Fix Favicon black background issue

mxlius:


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -710,7 +710,7 @@ public void run() {
                 return;
             }
             new FavIconUtils.SaveBitmapsTask(cacheDir, urls, icons, new UpdateHistoryWrapper(urls, handlerWeakReference),
-                    Bitmap.CompressFormat.JPEG, DimenUtils.JPEG_QUALITY).execute();
+                    Bitmap.CompressFormat.PNG, DimenUtils.JPEG_QUALITY).execute();
```
mTwTm: Let's also change DimenUtils.JPEG_QUALITY:
Make it's value = 0
rename to PNG_QUALITY_DONT_CARE

PR #2582, opened 2018-09-12
Ensure topsite is shown after all history upgrading is done.

### LINKED ISSUE #2581: Topsite is loaded when not all sites are migrated.
After upgrading, in the first launch Top site is still showing Captial letters rather than favicon.

mTwTm:
Fixes #2581

### DIFF
--- a/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
+++ b/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
@@ -862,15 +862,21 @@ public class HomeFragment extends LocaleAwareFragment implements TopSitesContrac
 
         @Override
         public void accept(List<String> fileUris) {
+            QueryHandler.AsyncUpdateListener listener = result -> {
+                Handler handler = handlerWeakReference.get();
+                if (handler == null) {
+                    return;
+                }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -862,15 +862,21 @@ private UpdateHistoryWrapper(List<String> urls, WeakReference<Handler> handlerWe
...
+                Message message = handler.obtainMessage(MSG_ID_REFRESH);
+                handler.dispatchMessage(message);
+            };
             for (int i = 0 ; i < fileUris.size() ; i++) {
-                BrowsingHistoryManager.updateHistory(null, urls.get(i), fileUris.get(i));
-            }
-            Handler handler = handlerWeakReference.get();
-            if (handler == null) {
-                return;
+                if (i == fileUris.size() - 1) {
```
mTwTm: Access to SQLite is [synchronous](https://www.androiddesignpatterns.com/2012/10/sqlite-contentprovider-thread-safety.html), so it is guaranteed that when the last entry is complete, all previous is also completed.

PR #2590, opened 2018-09-12
Don't write favicon into cache folder, create separate folder instead.

### LINKED ISSUE #2584: Favicon becomes all white after caches deleted
A simple str is to click clear cache manually in app info page.

mTwTm:
Fixes #2584

### DIFF
--- a/app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
@@ -32,9 +32,13 @@ import org.mozilla.focus.provider.QueryHandler;
 import org.mozilla.focus.site.SiteItemViewHolder;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.DimenUtils;
+import org.mozilla.focus.utils.FileUtils;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/history/HistoryItemAdapter.java
```diff
@@ -121,15 +125,17 @@ public void onResourceReady(Bitmap resource, Transition<? super Bitmap> transiti
...
-                    @Override
-                    public boolean onMenuItemClick(MenuItem menuItem) {
-                        if (menuItem.getItemId() == R.id.browsing_history_menu_delete) {
-                            BrowsingHistoryManager.getInstance().delete(item.getId(), HistoryItemAdapter.this);
-                            TelemetryWrapper.historyRemoveLink();
+                popupMenu.setOnMenuItemClickListener(menuItem -> {
+                    if (menuItem.getItemId() == R.id.browsing_history_menu_delete) {
+                        try {
+                            new FileUtils.DeleteFileThread(new File(new URI(item.getFavIconUri()))).start();
+                        } catch (URISyntaxException e) {
```
mxlius: How about let's separate the try/catch exception from new DeleteFileThread.
Can we reduce the usage of creating a new thread for each action?

Issue #1821, opened 2018-04-20
Experiment: Wifi Finder Offering (Landing Page Test)


joechengla:

weslyhuang:
need engineering study for some design input.

Jacklin-ux:
UX draft spec: https://drive.google.com/file/d/13-4IUB4DKoEi2apWJqHLTMZrsX3v7ELb/view?usp=sharing

chianger:
visual design will come soon. this is sibling to #1820

Pekochen:
wifi & vpn visual spec:
https://drive.google.com/open?id=1XULh6sqI2RxECra_aqvVAmUF9V9hqaiF
assets:
https://drive.google.com/open?id=1Fl0lnHb7ePeMJ1LsWvqq6UHwnICyPIfh
preview:
https://drive.google.com/open?id=1OeFAdt3wuyAYJShRtnOAfqiMUbGzLyxC

elin-moco:
@cnevinc My suggested event structure would be (in method/object/value format):
Click on Wifi Finder Button: click/home/wifi_finder
Click "Yes" on Wifi Finder Doorhanger: click/doorhanger/positive (extra: {source: wifi_finder})
Click "No" on Wifi Finder Doorhanger: click/doorhanger/negative (extra: {source: wifi_finder})
Dismissed Wifi Finder Doorhanger: click/doorhanger/dismiss (extra: {source: wifi_finder})
We'll need to put "wifi_finder" in extra for the doorhanger so that we can compare metrics in A/B testing.
I'll separate a dashboard to visualize the extra information.

cnevinc:
Please help answer below question for data review:

1) What questions will you answer with this data?

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:

* Establish baselines or measure changes in product or platform quality or performance.

* Provide information essential for advancing a business objective such as supporting OKRs.

* Determine whether a product or platform change has an effect on user or browser behavior.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?

4) Can current instrumentation answer these questions?

5) List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the Mozilla wiki.   

**Note that the data steward reviewing your request will characterize your data collection based on the highest (and most sensitive) category.**

<table>
  <tr>
[...]

Issue #1820, opened 2018-04-20
Experiment: VPN Offering


joechengla:

weslyhuang:
Two steps here:
1. wifi finder service: need to share cost with partner so plan to design a landing page to validate the user need.
2. VPN

weslyhuang:
need engineering study for some design input. (go with #1821's spike first)

Jacklin-ux:
UX draft spec: https://drive.google.com/file/d/13-4IUB4DKoEi2apWJqHLTMZrsX3v7ELb/view?usp=sharing

Pekochen:
wifi & vpn visual spec:
https://drive.google.com/open?id=1XULh6sqI2RxECra_aqvVAmUF9V9hqaiF
assets:
https://drive.google.com/open?id=1Fl0lnHb7ePeMJ1LsWvqq6UHwnICyPIfh
preview:
https://drive.google.com/open?id=1OeFAdt3wuyAYJShRtnOAfqiMUbGzLyxC

josephyu1110:
@Pekochen Please help to grant the access permission of specs to "joseph.yu@yarboroughtech.com". Thanks!

Pekochen:
@josephyu1110 Done : )

josephyu1110:
@Pekochen, I can access the second link now, but still cannot access the first and the third one(wifi & vpn visual spec and preview).

elin-moco:
@cnevinc My suggested event structure would be (in method/object/value format):
Click on VPN Button: click/home/vpn
Click "Yes" on VPN Doorhanger: click/doorhanger/positive (extra: {source: vpn})
Click "No" on VPN Doorhanger: click/doorhanger/negative (extra: {source: vpn})
Dismissed VPN Doorhanger: click/doorhanger/dismiss (extra: {source: vpn})
We'll need to put "vpn" in extra for the doorhanger so that we can compare metrics in A/B testing.
I'll separate a dashboard to visualize the extra information.

PR #2621, opened 2018-09-16
Implement WiFi finding & VPN service survey UI.

josephyu1110:
- #1820, #1821


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -903,27 +909,56 @@ private static void scheduleRefresh(Handler handler) {
...
-                textContent.setText(getString(R.string.exp_survey_thanks, "\uD83D\uDE00"));
-                btnYes.setVisibility(View.GONE);
-                btnNo.setVisibility(View.GONE);
-            });
-            btnNo.setOnClickListener(v3 -> {
-                textContent.setText(getString(R.string.exp_survey_thanks, "\uD83D\uDE00"));
-                btnYes.setVisibility(View.GONE);
-                btnNo.setVisibility(View.GONE);
-            });
+        synchronized (this) {
```
cnevinc: why do we need synchronized here?

josephyu1110: In case user tap the survey button very quickly, it might inflate a duplicate view. So I add a synchronized block to protect it.

app/src/main/java/org/mozilla/focus/home/FeatureSurveyViewHelper.kt
```diff
@@ -0,0 +1,102 @@
...
+import android.widget.Button
+import android.widget.TextView
+import org.mozilla.focus.R
+import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.utils.RemoteConfigConstants
+import org.mozilla.focus.utils.Settings
+
+class FeatureSurveyViewHelper internal constructor(private val context: Context, private val featureSurvey: RemoteConfigConstants.SURVEY) : View.OnClickListener {
+
+    private val dismissDelay: Long = 5000
```
cnevinc: nit: const and UPPER CASE

PR #2624, opened 2018-09-17
Fix verifyError: class org.robolectric.android.fakes.RoboMonitoringInstrumentation

cnevinc:
We see this error when running Roboletric tests

We should upgrade test_runner version as well when we upgrade roboletric version.


cnevinc:
I think we are already at the latest Roboletric version
https://github.com/mozilla-tw/Rocket/blob/master/buildSrc/src/main/java/Dependencies.kt#L27

PR #2633, opened 2018-09-18
Send the last ping before the user disable "Send Usage Data"

### LINKED ISSUE #2487: Know the number of users turn off "Send usage data"

cnevinc:
closes #2487

Confirmed with UX and product.
We'll send this ping using our best effort.
If the battery is low and the network is slow, the ping won't be sent.
This is because the Telemetry library uses JobScheduler to upload the ping.
And there's no control at our end. In the above situation, data loss is acceptable.


cnevinc:
This approach is meaningless. Close the pr

Issue #2137, opened 2018-07-04
Promote "set as default browser" on the first launch on preloads 


joechengla:

weslyhuang:
1. in order to track correctly, better to be preloaded as a system app
2. assuming preloaded as system app already. using the existing way (notification) for promotion.

question: notification style?

Jacklin-ux:
Proposal: after onboarding pages then display Rocket's home, show "Set as default browser" dropped-down notification (not hidden notification) to trigger users to set Rocket's as default browser.
My assumption is to be a preload browser, after users view onboarding pages, they've already known what's the benefit of the browser and known if they like this browser yes or not, so I think this display timing is suitable.

weslyhuang:
still keep the current one (enter setting twice) as long as Rocket is not set as default browser.

chianger:
to be verified.

PR #2673, opened 2018-09-25
Promote "set as default browser" on the first launch on preloads #2137

cnevinc:
We also allow Nightly and debug build to show the notification after first run finish


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
```diff
@@ -134,7 +134,7 @@ class FirstrunFragment : Fragment(), View.OnClickListener {
 
     private fun promoteSetDefaultBrowserIfPreload(): FirstrunFragment {
         // if it's a system app(preload), we'll like to promote set default browser when the user finish first run
-        if (isSystemApp() && !AppConstants.isReleaseBuild()) {
```
mTwTm: Maybe explain explicitly why we want non-release build to have this behavior too.

PR #2687, opened 2018-09-27
Upgrade to sdk 9.0

mTwTm:


app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +27,7 @@
         android:label="@string/app_name"
         android:supportsRtl="false"
         android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true"
```
walkingice: per the SDK documentation, the default value of `android:usesCleartextTraffic` is true, should we still explicitly specify the value?

mTwTm: It seems to not default true anymore since Android 9, see: https://developer.android.com/about/versions/pie/android-9.0-changes-28#framework-security-changes

PR #2695, opened 2018-09-28
Add Private Mode to menu panel, closes #2533

### LINKED ISSUE #2533: new Private Browsing tab from menu

mxlius:


app/src/main/res/layout/fragment_private_homescreen.xml
```diff
@@ -54,7 +54,7 @@
         android:id="@+id/pm_home_fake_input"
         style="@style/FakeInput"
         layout="@layout/fragment_homescreen_item_fake_input"
-        android:layout_width="match_parent"
+        android:layout_width="306dp"
```
mTwTm: Is this change per spec? If so, should we align this with normal browsing too? (Normal browsing is match_parent)

mxlius:
I'll fix the layout in another patch

Issue #2692, opened 2018-09-28
Handling Permissions Denials


joechengla:
Please only show the app settings button only after the user has set “Don’t ask again”. If the user has not set the permission to “Don’t ask again”, instead of taking the user to the settings, have the button request the permission again.

Jacklin-ux:
Proposal: https://files.slack.com/files-pri/T027LFU12-FD2AY91GQ/image.png

mTwTm:
Fixed in #2692

PR #2746, opened 2018-10-12
Handling Permissions denials #2692

cnevinc:
I have a strange situation here.
Maybe you've already seen it?

See https://github.com/mozilla-tw/Rocket/compare/nevin/permission?expand=1#diff-29976a057276bb0f502b51b61ba7dfe7R155

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -48,6 +48,8 @@ import android.webkit.WebView;
 import android.widget.CheckedTextView;
 import android.widget.FrameLayout;
 import android.widget.RelativeLayout;
+import android.widget.Toast;
+
 
 import org.mozilla.focus.R;
 import org.mozilla.focus.download.EnqueueDownloadTask;
@@ -335,6 +337,11 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
                     default:
                         throw new IllegalArgumentException("Unknown Action");
                 }
+            }
+
+            @Override
+            public void permissionDeniedToast() {
+                Toast.makeText(getContext(), R.string.permission_toast_location_deny, Toast.LENGTH_LONG).show();
             }
         });
     }
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -321,6 +321,16 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
                 }
             }
 
+            private int getPermissionDeniedToastString(int actionId) {
+                if (actionId == ACTION_DOWNLOAD || actionId == ACTION_PICK_FILE || actionId == ACTION_CAPTURE) {
+                    return R.string.permission_toast_storage_deny;
+                } else if (actionId == ACTION_GEO_LOCATION) {
+                    return R.string.permission_toast_location_deny;
+                } else {
+                    throw new IllegalArgumentException("Unknown Action");
+                }
+            }
+
             @Override
             public void requestPermissions(int actionId) {
                 switch (actionId) {
@@ -341,7 +351,7 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
 
             @Override
             public void permissionDeniedToast(int actionId) {
-                Toast.makeText(getContext(), getAskAgainSnackBarString(actionId ), Toast.LENGTH_LONG).show();
+                Toast.makeText(getContext(), getPermissionDeniedToastString(actionId), Toast.LENGTH_LONG).show();
             }
         });
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -141,13 +146,18 @@ public void onRequestPermissionsResult(Context context, int requestCode, @NonNul
...
         if (requestCode == actionId) {
-            setPermissionAsked(context, permissions[0]);
+
             if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                 permissionHandle.doActionGranted(permission, actionId, params);
                 clearAction();
             } else {
-                showSnackBar();
+                showToast();
+                // I'm not sure why it's so. This method already on Main thread.
```
mTwTm: You need to do permissionNotGranted() here, note that we do call this in showSnackBar() but not in showToast(). Sorry for writing bad interfaces :P

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -334,6 +335,11 @@ public void requestPermissions(int actionId) {
                         throw new IllegalArgumentException("Unknown Action");
                 }
             }
+
+            @Override
+            public void permissionDeniedToast() {
```
mTwTm: We have different permissions, location and storage. Pass actionId to this interface, see Snackbar makeAskAgainSnackBar(int actionId);

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -141,13 +146,18 @@ public void onRequestPermissionsResult(Context context, int requestCode, @NonNul
...
         }
         if (requestCode == actionId) {
-            setPermissionAsked(context, permissions[0]);
+
             if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                 permissionHandle.doActionGranted(permission, actionId, params);
                 clearAction();
             } else {
-                showSnackBar();
+                showToast();
```
mTwTm: Tried if we don't reject, it's still incorrect. Maybe we should first show the dialog first and only if the user wants to give the site permission then we ask for device permission. This is how Firefox and Chrome handle this.

app/src/main/java/org/mozilla/focus/permission/PermissionHandler.java
```diff
@@ -141,13 +146,18 @@ public void onRequestPermissionsResult(Context context, int requestCode, @NonNul
...
+                // I'm not sure why it's so. This method already on Main thread.
+                // But if I don't do this, webview will keeps requesting for permission.
+                // See https://github.com/mozilla-tw/Rocket/blob/765f6a1ddbc2b9058813e930f63c62a9797c5fa0/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebChromeClient.java#L126
+                ThreadUtils.postToMainThread(this::permissionNotGranted);
             }
+            setPermissionAsked(context, permissions[0]);
```
mTwTm: Do we need this?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
cnevinc: fixed

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -336,6 +338,11 @@ public void requestPermissions(int actionId) {
                         throw new IllegalArgumentException("Unknown Action");
                 }
             }
+
+            @Override
+            public void permissionDeniedToast(int actionId) {
+                Toast.makeText(getContext(), getAskAgainSnackBarString(actionId ), Toast.LENGTH_LONG).show();
```
mTwTm: I think this is not what the spec describes?

cnevinc: fixed

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -336,6 +348,11 @@ public void requestPermissions(int actionId) {
                         throw new IllegalArgumentException("Unknown Action");
                 }
             }
+
+            @Override
+            public void permissionDeniedToast(int actionId) {
+                Toast.makeText(getContext(), getPermissionDeniedToastString(actionId ), Toast.LENGTH_LONG).show();
```
mTwTm: NIT: extra space.

app/src/main/res/values/strings.xml
```diff
@@ -250,6 +250,8 @@
     <string name="permission_toast_location">Location not available. Give permission to access your location.</string>
     <string name="permission_dialog_msg_storage">Missing permissions. Give Firefox Rocket permission to access your location.</string>
     <string name="permission_toast_storage">Storage not available. Give permission to store photos and files on your device.</string>
+    <string name="location_permission_is_denied">Location permission is denied.</string>
```
mTwTm: We already have: permission_toast_location_deny

app/src/main/res/values/strings.xml
```diff
@@ -250,6 +250,8 @@
...
+    <string name="storage_permission_is_denied">Storage permission is denied.</string>
```
mTwTm: Same here.

Issue #2750, opened 2018-10-13
(Night mode) Strange brightness when drag the thumb to the end of left side


Jacklin-ux:
Enable night mode then drag the thumb on the slider. When drag the thumb to the end of the left side, the brightness will be using system brightness instead of the darkest brightness
![screenshot oct 15 2018 10_34_53 am](https://user-images.githubusercontent.com/22672569/46927061-3c32bd80-d066-11e8-825e-e8482a3aff75.png)

weslyhuang:
Joseph can't repro it. Can QA help on it? tks

vanceyen:
Seems like this bug only happens on Google native Android phone(e.g. Pixel 2), also @mxlius is working on this one already ?

mxlius:
Fixed in #2752

PR #2752, opened 2018-10-15
Update brightness default behavoiur, closes #2751

### LINKED ISSUE #2751: (Night mode) Brightness range is from 0 to 50% and the thumb's default location is at 1/4
The brightness range which users can adjust on the bar is from 0 to 50% of system brightness. And the default location of the thumb is at 1/4 which is 12.5%

mxlius:


mxlius:
Closes #2750

Issue #1910, opened 2018-05-09
Night mode


Jacklin-ux:
Accounting to "Report for Firefox Rocket" user feedback, many users would like to have night mode in our Rocket.

weslyhuang:
plan to get BJ's support for this feature. UX to provide the icon for the first version.

Pekochen:
Night mode visual spec:
https://drive.google.com/open?id=1hVFl_GxvzbSRl51xXUpo5XIwMf2Mbcpt

Pekochen:
According to engineer's request, add the previous spec of normal mode
https://drive.google.com/open?id=12EfqIPRc_uZJjkoCHdO0ApbFyDDXTf55

weslyhuang:
use prototype to check first. ship or not depends on the damage situation of main top 20~25 sites (focus on "article/shopping" sites) of target countries.

cnevinc:
Please download the app-focus-webkit-firebase.apk in deploy.zip.
here's the link
https://app.bitrise.io/build/c370f1645728d2d5

weslyhuang:
possible alternative: don't play with color but allow users to adjust brightness.

weslyhuang:
1. a test apk with adjustable lightness
2. UX try and decide the lightness levels
3. design UX (e.g., scroll bar) for it

Jacklin-ux:
UX draft spec: https://drive.google.com/file/d/1zPToaAZmnt1BfqKCo0oX8H_GNb0_H9Wr/view?usp=sharing

Pekochen:
update night mode visual spec and add brightness adjust control panel
https://drive.google.com/open?id=1vkONwjHSwk-gZglY4lekVfde2_zZjsMc
assets:
https://drive.google.com/open?id=16hT0x5xOQbmps5qElMp6UiRFvn-7YdIT
preview:
https://drive.google.com/open?id=1YKEr-qMtvC0LPAjUWht2OVeLYpuSseEa

chianger:
we have prototype for adjusting lightness in place, for parameter tuning. 
still need to figure out use cases for "exiting"  night mode so this keeps in design.

chianger:
correction for above comment:
let's take that auto-exit night mode use case as a follow up feature.

chianger:
size-wanted because we're not sure what telemetry needed for this feature. that's a dependency for the implementation.
Telemetry-wanted for @joechengla @elin-moco to comment on the telemetry
[...]

PR #2764, opened 2018-10-16
Add telemetry for night mode.

josephyu1110:
- #1910

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/AdjustBrightnessDialog.kt
+++ b/app/src/main/java/org/mozilla/focus/widget/AdjustBrightnessDialog.kt
@@ -1,10 +1,13 @@
 package org.mozilla.focus.widget
 
+import android.content.Context
+import android.content.Intent
 import android.os.Bundle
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/widget/AdjustBrightnessDialog.kt
```diff
@@ -27,6 +30,29 @@ class AdjustBrightnessDialog : BaseActivity() {
         }
     }
 
+    object Intents {
```
cnevinc: Sorry, my bad. I didn't think through enough.
Base on the simple responsibility principle, AdjustBrightnessDialog shouldn't contain utility method and both dialog logic. We should be this object standalone in the package.
Maybe move the dialog and this object to a new package in "org.mozilla.rocket.nightmode"

josephyu1110: Updated.

PR #2771, opened 2018-10-17
Add SpotlightOnboarding to NightMode

mxlius:


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -397,10 +400,13 @@ private void showMenu() {
     }
 
     private void updateMenu() {
+
         final boolean isMyShotUnreadEnabled = AppConfigWrapper.getMyshotUnreadEnabled(this);
         final boolean showUnread = isMyShotUnreadEnabled && Settings.getInstance(this).hasUnreadMyShot();
         myshotIndicator.setVisibility(showUnread ? View.VISIBLE : View.GONE);
         if (pendingMyShotOnBoarding) {
+            pendingMyShotOnBoarding = false;
+            setShowNightModeSpotlight(false);
```
cnevinc: I feel we have a need to write tests to protect the logic here. I've created #2785 for this.

app/src/main/java/org/mozilla/focus/utils/Settings.java
```diff
@@ -76,6 +76,16 @@ public void setNightMode(boolean enable) {
         preferences.edit().putBoolean(key, enable).apply();
     }
 
+    public void setNightModeSpotlight(boolean enabled) {
+        final String key = getPreferenceKey(R.string.pref_key_night_mode_brightness_dirty);
+        preferences.edit().putBoolean(key, enabled).apply();
+    }
+
+    public boolean shouldShowSpotlight() {
```
cnevinc:
```suggestion
    public boolean shouldShowNightModeSpotlight() {
```

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -704,6 +757,14 @@ private boolean isNightModeEnabled() {
         return Settings.getInstance(this).isNightModeEnable();
     }
 
+    private boolean shouldShowNightModeSpotlight(){
+        return Settings.getInstance(this).shouldShowSpotlight();
```
cnevinc:
```suggestion
        return Settings.getInstance(this). shouldShowNightModeSpotlight();
```

app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
```diff
@@ -206,6 +206,59 @@ public static void showPrivacyPolicyUpdateNotification(Context context) {
         NewFeatureNotice.getInstance(context).setPrivacyPolicyUpdateNoticeDidShow();
     }
 
+    public static Dialog showSpotlight(@NonNull final Activity activity, @NonNull final View targetView, @NonNull DialogInterface.OnCancelListener onCancelListener, int messageId){
```
cnevinc: Can we re-use the code in showMyShotOnBoarding() and rename it to showSpotlight?

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -463,6 +480,41 @@ private void dismissAllMenus() {
         }
     }
 
+    OnLongClickListener onLongClickListener = new OnLongClickListener() {
+        @Override
+        public boolean onLongClick(View v) {
+            menu.cancel();
+            switch (v.getId()) {
+                case R.id.menu_night_mode:
+                    final Settings settings = Settings.getInstance(MainActivity.this);
```
cnevinc: can we just call onNightModeClicked() in MainActivity?

app/src/main/java/org/mozilla/focus/utils/DialogUtils.java
mxlius: I think it would be better if we separate the reuse clean up to another patch

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
mxlius: Agreed, I cleaned up the flow in onNightModeClicked() to reduce duplicate code.

PR #2786, opened 2018-10-19
Change app icon to lite version

mTwTm:


mTwTm:
I thought we have a conclusion that we'll like to follow whatever visual provides? But yeah given the logo is within the same usage, I think renaming is more aligning to that spirit. I'll fix later.

PR #2796, opened 2018-10-22
Move Night Mode to module

cnevinc:


app/src/main/java/org/mozilla/rocket/nightmode/AdjustBrightnessDialog.kt
```diff
@@ -30,29 +29,6 @@ class AdjustBrightnessDialog : BaseActivity() {
         }
     }
 
-    object Intents {
```
cnevinc: Please put them in nightmode module. create a kt file, put this object there.
IntentUtils is exploding now.

components/feature/night-mode/src/main/java/org/mozilla/rocket/nightmode/themed/NightTheme.kt
```diff
@@ -0,0 +1,28 @@
...
+            }
+            notifyRefreshDrawableState()
+        }
+    }
+
+    fun isNightTheme(): Boolean {
+        return themeState and ThemedWidgetUtils.ThemeState.NIGHT.flag != 0
+    }
+
+    override fun getThemeStateSize(): Int {
```
mTwTm: I'm not sure if we really need to expose this. Maybe just let outside users use getThemeDrawableState().size?

components/feature/night-mode/src/main/java/org/mozilla/rocket/nightmode/themed/NightTheme.kt
```diff
@@ -0,0 +1,28 @@
...
+        return ThemedWidgetUtils.STATE_NIGHT_MODE.size
+    }
+
+    override fun getThemeDrawableState(): IntArray {
```
mTwTm: Just asking: So, for NightPrivateTheme, it will implement both NightTheme and PrivateTheme and have a static intArray member which merges ThemedWidgetUtils.STATE_NIGHT_MODE and hemedWidgetUtils.STATE_PRIVATE_MODE?

components/feature/night-mode/src/main/java/org/mozilla/rocket/nightmode/themed/NightTheme.kt
josephyu1110: Good point. I haven't noticed that!

components/feature/night-mode/src/main/java/org/mozilla/rocket/nightmode/themed/NightTheme.kt
josephyu1110: Yes, it should be like this.
`override fun getThemeDrawableState(): IntArray {
        return when {
            isNightTheme() && isPrivateTheme() -> super<NightTheme>.getThemeDrawableState().plus(super<PrivateTheme>.getThemeDrawableState())
            isNightTheme() -> super<NightTheme>.getThemeDrawableState()
            isPrivateTheme() -> super<PrivateTheme>.getThemeDrawableState()
            else -> intArrayOf(0)
        }
    }`

mTwTm: Sounds good to me, ready to go.

app/src/test/java/org/mozilla/rocket/nightmode/NightModeTestUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.robolectric.Robolectric
+import org.robolectric.RuntimeEnvironment
+
+import org.junit.Assert.assertTrue
+
+internal object NightModeTestUtils {
+
+    private val TARGET_BRIGHTNESS = 30.0f
+
+    fun <T : Activity> testActivityBrightness(classname: Class<T>): T {
```
cnevinc: This more like UI test. Not unit test. Please remove the tests and let's land the code first. We'll discuss the unit test later when we have time. Thank you!

josephyu1110: It's done.

PR #2824, opened 2018-10-24
tabs: fix closing tab crash issue

walkingice:
when web page called `window.close()` by itself, which triggers
`Session.ChromeClient.onCloseWindow`. In the implementation, which
invokes `session.notifyObservers`

SessionManager is one of observers, in its callback function it calls
`session.destroy()` which cause concurrent-modification exception.

This commit schedule the removing(destroy) to next round of execution.


components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
```diff
@@ -39,6 +39,7 @@ class SessionManager @JvmOverloads constructor(
 
 
     private val sessions = LinkedList<Session>()
+    private val removed = LinkedList<Session>()
```
mTwTm: Are we using this at all?

walkingice: No. I should remove it

PR #2839, opened 2018-10-27
Ac/involve engine session

walkingice:
This PR involves TabViewEngineSession and TabViewEngineObserver.

Rocket has its own WebView abstraction - TabView. Before we involves EngineView from Android-Components, add this abstraction then we can make Session/SessionManager more like what is existing in AC.

This PR also simplify interfaces of Session. Now it becomes only a state-holder, just like the Session of AC.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -210,6 +210,11 @@ dependencies {
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${Versions.kotlin}"
 
+    // We didn't use CustomTabs so far. This is a build hack to force Android-Components to use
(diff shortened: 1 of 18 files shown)


walkingice:
Android-Component is using **27.1.1** support library which is different from what Rocket is using (28.0.0). Bitrise failed on `All com.android.support libraries must use the exact same version specification`

ref: https://github.com/mozilla-mobile/android-components/issues/404

walkingice:
I had added comments to TabViewEngineSession by following Nevin's suggestion. Wish we could remove them as soon as possible.

PR #2853, opened 2018-10-31
Ac/involve_ac_session

walkingice:
This PR added SessionFeature which encapsulate details of using Session/SessionManager for BrowserFragment. It also involve Android-Component's Session into Rocket's code base.


cnevinc:
Rebasing, and need to make sure we don't miss the fixes we've applied since last month.

mTwTm:
I see a lot of conflicts. Is this still active?

PR #2862, opened 2018-11-02
Upgrade oss-licenses-plugin to avoid licenses list checkin to source …

mxlius:
…tree


mxlius:
The list is generated during runtime. Do we intend to provide the list with an URL to external reference?

cnevinc:
Please merge this asap :P since it helps a lot!

PR #2863, opened 2018-11-02
[third-party][glide] Fix build warning of targetSdkVersion with Android Gradle Plugin 3.2.0

mxlius:
`targetSdkVersion` should not place in AndroidManifest.xml, put it into `build.gradle` instead.


mTwTm:
Should be subsampling-scale-image-view for this case instead of Glide, though :P

PR #2882, opened 2018-11-07
Use referral link instead

cnevinc:


chianger:
To enable VPN recommender we would need this URL also remote configurable.

app/src/main/java/org/mozilla/focus/home/FeatureSurveyViewHelper.kt
```diff
@@ -82,8 +84,8 @@ class FeatureSurveyViewHelper internal constructor(private val context: Context,
                     // Dismiss view immediately
                     parentView.removeView(rootView)
                     isViewInit = false
-                    // VPN app is installed, go to google play store
-                    IntentUtils.goToPlayStore(context, Constants.PACKAGE_EXPRESS_VPN)
+                    // VPN app is installed, open Play Store in a new tab
```
mxlius: I guess this comment should be "VPN app is NOT installed", so we open play store in a new tab?

firebase/src/firebase/java/org/mozilla/focus/utils/FirebaseWrapper.java
```diff
@@ -74,12 +74,12 @@
 
 
     // get Remote Config string
-    static String getRcString(@NonNull Context context, @NonNull String key) {
```
mTwTm: You might need to change the one in firebase_no_op/.../FirebaseWrapper.java as well.

cnevinc:
Actually, this is where we put the default value
https://github.com/mozilla-tw/Rocket/blob/e193c2a13f8cda884176129f9f1e6b427c6ce3da/app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java#L251
Cause we don't want to open another interface to let client code setting the default value.

PR #2899, opened 2018-11-13
Glue SessionManager and TabViewEngineSession.windowClient

### LINKED ISSUE #2724: User cannot choose file to upload
Description
User cannot upload file

STR
1. Launch Rocket
2. Visit a website (i.e. encodable.com/uploaddemo) that can test file upload
3. Upload a file

Expected result
User can upload files

Actual result
User cannot choose file to upload

Clip
https://drive.google.com/open?id=1BQpEAwIfijvLbVW08IW4nQDT-qUzG8ax

Device and Rocket Info
Rocket 3.4.0(7421)
Device: Nexus 5
WebView: 69.0.3497.100
Android: 6.0.1

Working version: Rocket 3.3.0 (7339)
Clip
https://drive.google.com/open?id=1iU2DKQ9wksWUcUlHInIiFmjr1yb0Whwj

Device: Nexus 5
WebView: 69.0.3497.100
Android: 6.0.1

mTwTm:
A draft to fix #2724 (reprise)

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
@@ -12,6 +12,8 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.CookieManager
+import android.webkit.ValueCallback
+import android.webkit.WebChromeClient
 import android.webkit.WebView
 import android.widget.ImageButton
 import android.widget.ImageView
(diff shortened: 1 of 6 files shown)


mTwTm:
Add @cnevinc to unblock 1.0.1. Please revert/fix this if this is not the desired implementation.

app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -229,6 +232,19 @@ class BrowserFragment : LocaleAwareFragment(),
     }
 
     class Observer(val fragment: BrowserFragment) : SessionManager.Observer, Session.Observer {
+        override fun updateFailingUrl(url: String?, updateFromError: Boolean) {
```
cnevinc: I feel like we can have one default observer here since they all do nothing

mTwTm: I tried to use java8 default methods but it needs some extra flags. I think a default observer is not feasible here given that there are interfaces that extends client so that would make the code even harder to read.

chianger:
this also fix #2902

PR #2906, opened 2018-11-14
Use network security config to allow DownloadManager download http URL

mxlius:


app/src/main/AndroidManifest.xml
```diff
@@ -28,6 +28,7 @@
         android:supportsRtl="false"
         android:theme="@style/AppTheme"
         android:usesCleartextTraffic="true"
```
mTwTm: Consider removing this for simplicity.

PR #2923, opened 2018-11-19
Nevin/determine private mode

### LINKED ISSUE #2542: [Google Play] Crash determine private or normal browsing process.
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
       at org.mozilla.focus.FocusApplication.isInPrivateProcess(FocusApplication.kt:73)
       at org.mozilla.focus.FocusApplication.getCacheDir(FocusApplication.kt:28)
       at org.mozilla.focus.home.HomeFragment.readFromCache(HomeFragment.java:247)
       at org.mozilla.focus.home.HomeFragment.setUpBanner(HomeFragment.java:208)
       at org.mozilla.focus.home.HomeFragment.access$400(HomeFragment.java:78)
       at org.mozilla.focus.home.HomeFragment$2.onReceive(HomeFragment.java:346)
       at android.support.v4.content.LocalBroadcastManager.executePendingBroadcasts(LocalBroadcastManager.java:311)
       at android.support.v4.content.LocalBroadcastManager.access$000(LocalBroadcastManager.java:47)
       at android.support.v4.content.LocalBroadcastManager$1.handleMessage(LocalBroadcastManager.java:120)
       at android.os.Handler.dispatchMessage(Handler.java:111)
       at android.os.Looper.loop(Looper.java:194)
       at android.app.ActivityThread.main(ActivityThread.java:5637)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:959)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:754)

cnevinc:
closes #2542

FocusApplication need to know if we are in private mode process, so it
can know when to override getCacheDir to provide the cache folder for
webview/chromium

We use process's name to see if it contains "private_mode" using
acitivityManager.getRunningAppProcesses(). But this method
sometimes return null list so a NullPointerException will be thrown
if we want to use it to check if we are in private mode.

We want to use a new way to determine if we are in prviate mode
process. We want to use the existance of PrivateModeActivity here.

But Telemetry.init() call in FocsApplication will make this hard cause
it make getCacheDir called much more early than PrivateModeActivity is
created. This is because it checked the webview's version in
TelemtryWrapper.init() and call the constructor of WebView.

This makes the chromium code call context.getApplicationContext().getCacheDir() too
early even before the acitvity is created.

Our solution here is to remove the webview version check. But maybe we
can put the version check after the first webView is created and save it
to a shared preference. At the next launch, we just read the pref but we
don't create a webview again just to check the version.

The assumption here is that as long as there's PrivateModateActivity,
this application is in Private Mode process.
This assumption won't hold true if private mode lives in the same
process of normal mode. But this is the design right now.

We should also test if the activity is killed by the system, will this
still work? I assume it will cause there'll never be a
PrivateModeActivity in normal mode.
Need more testing so this can be landed.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/FocusApplication.kt
+++ b/app/src/main/java/org/mozilla/focus/FocusApplication.kt
@@ -5,8 +5,8 @@
 
 package org.mozilla.focus
 
-import android.app.ActivityManager
-import android.content.Context
+import android.app.Activity
+import android.os.Bundle
 import android.preference.PreferenceManager
 import com.squareup.leakcanary.LeakCanary
 import org.mozilla.focus.download.DownloadInfoManager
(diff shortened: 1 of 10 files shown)


cnevinc:
There is a bunch of methods to get the correct process name. [see this](https://hk.saowen.com/a/fd9e31cc3ff0c98fc8ea1eb4207fb059509b37cfae6c5846ab105c91067f38e7)
But I feel they are all hacky. So I choose to use PrivateModeActivity's existence to determine if we are in private mode process.

PR #2925, opened 2018-11-19
Check Webview.canGoback before doing Webview.goBack

mTwTm:
The webview.goBack here refers to not only goback itself but also
the WebHistoryItem we're trying to get which causes a crash.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -880,7 +880,10 @@ public class BrowserFragment extends LocaleAwareFragment implements View.OnClick
         final Session currentTab = sessionManager.getFocusSession();
         if (currentTab != null) {
             final TabView current = currentTab.getEngineSession().getTabView();
-            if (current == null) {
+            // The Session.canGoBack property is mainly for UI display purpose and is only sampled
+            // at onNavigationStateChange which is called at onPageFinished, onPageStarted and
+            // onReceivedTitle. We do some sanity check here.
+            if (current == null || !current.canGoBack()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -880,7 +880,10 @@ public void goBack() {
         final Session currentTab = sessionManager.getFocusSession();
         if (currentTab != null) {
             final TabView current = currentTab.getEngineSession().getTabView();
-            if (current == null) {
+            // The Session.canGoBack property is mainly for UI display purpose and is only sampled
+            // at onNavigationStateChange which is called at onPageFinished, onPageStarted and
+            // onReceivedTitle. We do some sanity check here.
+            if (current == null || !current.canGoBack()) {
```
mxlius: The implementation looks different from the one `BrowserFragment.canGoBack()` checked in `onBackPressed`
Should we remove that and keep this only?

We handled `canGoBack`, how about `canGoForward`?

mTwTm: We don't currently have a BrowserFragment.canGoBack() visual hint. (On the other hand, we do have one for canGoForward) If we are to remove the BrowserFragment.canGoBack() altogether, we better also remove the Session.canGoBack() altogether since BrowserFragment.canGoBack() is the only place that calls Session.canGoBack(). I tend to believe that the parity between canGoBack() and canGoForward actually makes the logic more maintainable so I'll keep them as is.

I can for sure do the same for current.goForward as well but given this patch is going to be a hotfix, I tend to do it in another pull request to minimize the change.

PR #2927, opened 2018-11-19
Update action to category for TelemetryDoc annotation

cnevinc:


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -59,7 +59,7 @@ object TelemetryWrapper {
     private const val SEARCHDISMISS_TELEMETRY_VERSION = "2"
 
     internal object Category {
-        const val ACTION = "action"
+        const val ACTION = "category"
```
mTwTm: I don't think we should change this too?

PR #2928, opened 2018-11-20
Enable Android App Bundle

mxlius:


app/build.gradle
```diff
@@ -318,7 +330,7 @@ clean.doLast {
 
 tasks.whenTaskAdded { task ->
     def name = task.name
-    if (name.contains("generate") && name.contains("Config") && name.contains("Webkit")) {
+    if (name.contains("generate") && name.contains("Webkit") && name.contains("Resources")) {
```
mTwTm: This will match: generateFocusWebkitCoverageResources, mergeFocusWebkitCoverageResources, processFocusWebkitCoverageResources, transformResourcesWithMergeJavaResForFocusWebkitCoverage, transformClassesAndResourcesWithProguardForPreviewWebkitRelease, i.e., 4/5 times fo each flavor. Are you sure we want to do this?

mTwTm: Oops there's actually a generate filter so should be fine.

PR #2931, opened 2018-11-20
Notify url changes in onReceivedTitle as well.

### LINKED ISSUE #2930: Url does not always change after tapping a link.
**Describe the bug**
For instance, after tapping a video in Youtube, The URL bar should be updated to youtube.com/watch?v=<hash>

mTwTm:
This is an old hack we used before 72262855 and it was
erronously removed in that commit.

Fixes #2930


mTwTm:
Sorry, should actually be 5662ba90 but it is not building so the first buildable version is 7226285

PR #2933, opened 2018-11-20
Make SpotlightDialog reusable

mxlius:


mTwTm:
Consider landing this to prevent bitrot.

PR #2947, opened 2018-11-21
Make sure when open from external App, the new tab will be at the end.

### LINKED ISSUE #2904: New tab is inserted at the top when open from external
**Steps To Reproduce**
1. Search in Firefox Lite or open external link in Firefox Lite

**Expected behavior**
2. A new tab should be inserted at the bottom of the tab list

**Actual behavior**
2. A new tab is inserted at the top of tab list

Reproduced on multiple devices.

**Reproduce rate**
Always

mTwTm:
Fixes #2904

### DIFF
--- a/components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
+++ b/components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
@@ -75,39 +75,14 @@ class SessionManager @JvmOverloads constructor(
         return ArrayList(sessions)
     }
 
-    /**
-     * To append sessions from a list of TabModel. If the specified focusTabId exists, the tab associate
-     * to the id will be focused, otherwise no tab will be focused.
-     *
-     *
-     * This is asynchronous call.
-     * TODO: make it asynchronous
-     *
-     * @param sessionList
-     */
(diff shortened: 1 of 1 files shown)


components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
```diff
@@ -75,39 +75,14 @@ class SessionManager @JvmOverloads constructor(
...
-
-        if (this.sessions.size > 0 && this.sessions.size == sessionList.size) {
-            focusRef = WeakReference<Session>(getTab(focusTabId))
-        }
-
-        notifyObservers { onSessionCountChanged(sessions.size) }
-    }
-
     fun restore(states: List<SessionWithState>, focusTabId: String?) {
+        // If Lite opened from external, 0 could be already occupied and should b the last item after restore.
```
cnevinc: nit: s/b/be

mTwTm: I can find where you mean by be but what's s and b?

mTwTm: Ah it's should be, thanks!

PR #2949, opened 2018-11-21
Try send adjust event ping

cnevinc:


app/src/preview/java/org/mozilla/focus/utils/AdjustHelper.java
```diff
@@ -5,10 +5,78 @@
...
+            throw new IllegalStateException("No adjust token defined for release build");
+        }
+
+        if (!TelemetryWrapper.isTelemetryEnabled(application)) {
+            return;
+        }
+
+        final AdjustConfig config = new AdjustConfig(application,
+                BuildConfig.ADJUST_TOKEN,
+                AdjustConfig.ENVIRONMENT_SANDBOX,   //sandbox only
```
mxlius: Looks like we copy the file content from the release one, can we reuse the same file with this parameter different in different flavor?

app/src/main/java/org/mozilla/focus/telemetry/AdjustConstant.kt
```diff
@@ -0,0 +1,3 @@
+package org.mozilla.focus.telemetry
+
+const val EVENT_TAKE_SCREENSHOT = "ky29gk"
```
mxlius: Is this randomly generated id? do we need to sync this with @elin-moco ?

cnevinc: @elin-moco give me this key. it's generated by Adjust backend. We can't change it.

app/src/preview/java/org/mozilla/focus/utils/AdjustHelper.java
cnevinc: yes. see todo in https://github.com/mozilla-tw/Rocket/pull/2949#issue-232637917

cnevinc: fixed

PR #2953, opened 2018-11-22
Restore stored webviewstate properly

### LINKED ISSUE #2909: Tabs can't be resumed after Rocket have gone to background but not killed
**Describe the bug**
Not always reproducible. The existing tabs become blank and reload doesn't help.

**Steps To Reproduce**
1. Launch Firefox Lite and open a few tabs.
2. Launch another app. For instance Chrome, and select text to search in firefox lite
3. Firefox Lite is then brought up to the front. 
4. Tap on tab list view and select any tab other than the newly created one

**Expected behavior**
The page display correctly

**Actual behavior**
The page is all blank, though the URL bar seems fine. 
Reload doesn't help.

**Screenshots/videos**
If applicable, add screenshots/videos to help explain your problem.

 - Device: PixelXL
 - OS: Android 8.1.0
 - Webview 70.0.3538.80
 - Version 1.0.1(8154)

**Reproduce rate**
2/6

Relevant to #2904

### LINKED ISSUE #2951: Tab history is discarded
**Describe the bug**
Tab history is discarded

**Steps To Reproduce**
1. Open a tab
2. browse pages, go to next page, and so on.
3. Kill Lite
4. Open lite
5. Press back

**Expected behavior**
The last page visited is opened.

**Actual behavior**
Home is opened because the history is discarded.

Only on 1.0.1

mTwTm:
In ab41654, we have changed the way we restore webviewstates. The deserialized states are
stored in a temporary storage SessionWithState. However, the webview state failed to be
passed over to the created engineSession. This commit fixes this missed delegation.

Fixes #2909, Fixes #2951

### DIFF
--- a/components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
+++ b/components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
@@ -82,6 +82,7 @@ class SessionManager @JvmOverloads constructor(
         for (state in states) {
             if (state.session.isValid()) {
                 getOrCreateEngineSession(state.session).let { link(state.session, it) }
+                state.session.engineSession?.webViewState = state.engineSession?.webViewState
                 this.sessions.add(insertPos++, state.session)
             }
         }


components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/SessionManager.kt
```diff
@@ -82,6 +82,7 @@ class SessionManager @JvmOverloads constructor(
         for (state in states) {
             if (state.session.isValid()) {
                 getOrCreateEngineSession(state.session).let { link(state.session, it) }
+                state.session.engineSession!!.webViewState = state.engineSession?.webViewState
```
cnevinc: state.session.engineSession could be null.
Now we have some !! in the codebase and we lost Kotlin's null-safety benefit. Maybe we should think of a better way to fix it.

mTwTm: Yeah I thought it's impossible but I'm wrong, thanks!

PR #2962, opened 2018-11-26
Don't send events to Adjust when user disables "Send Usage Data"

cnevinc:


app/src/release/java/org/mozilla/focus/utils/AdjustHelper.java
```diff
@@ -19,7 +19,12 @@
 import org.mozilla.focus.FocusApplication;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 
+import java.lang.ref.WeakReference;
+
 public class AdjustHelper {
+
+    private static WeakReference<FocusApplication> weakApplication;
```
mxlius: `Application` instance lifecycle is like static, it doesn't make sense to have it in a WeakReference

PR #2964, opened 2018-11-26
Support MixedContent in Settings, Closes #954

### LINKED ISSUE #954: support mixed contents
https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content

mxlius:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -575,6 +575,15 @@ public void setContentBlockingEnabled(boolean enabled) {
         }
     }
 
+    public void setAllowMixedContent(boolean allowMixedContent) {
+        for (final Session session : sessionManager.getTabs()) {
```
mTwTm: NIT: Not sure if we should keep those todos. They are fairly old, though.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -847,6 +847,12 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
             }
             setMenuButtonSelected(R.id.menu_blockimg, blockingImages);
+        } else if (this.getResources().getString(R.string.pref_key_allow_mixed_content).equals(key)) {
+            boolean allowMixedContent = Settings.getInstance(this).getAllowMixedContent();
+            BrowserFragment browserFragment = getBrowserFragment();
+            if (browserFragment != null) {
+                browserFragment.setAllowMixedContent(allowMixedContent);
+            }
         }
         // For turbo mode, a automatic refresh is done when we disable block image.
```
mxlius: https://github.com/mozilla-tw/Rocket/pull/2964/files/0c8cb9627501b40ef5e35bbb14a39153df1b9352#r236575804

I guess you mean this line of comment?
I think yes, we should remove this comment or at least move/merge it to first if statement.

mTwTm: I actually mean
```
        // TODO: Better if we can move this logic to some setting-like classes, and provider interface
        // for configuring blocking function of each tab.
```

mTwTm: Your statement is true too.

mTwTm:
The build is still failing for unknown reason, I'm not sure it is because we have treat kotlin warnings as errors but if that is the case I'm not sure why this is not caught in some earlier prs.

PR #2980, opened 2018-11-29
Make Remote Config remote value locale aware. closes #2794

### LINKED ISSUE #2794: Change back Language to English doesn't work in Give Feedback area
**Describe the bug**
Change back Language to English doesn't work in Give Feedback area

**Steps To Reproduce**
1. Launch Rocket
2. Tap on Menu and then Settings
3. Tap on Language and choose "Indonesian"
4.  Back to Settings, tap on Language and choose "English"
6. Tap on Give feedback

**Expected behavior**
Showed in English 
**Actual behavior**
Showed in Indonesia.
**Screenshots/videos**
https://drive.google.com/open?id=1WH1ezZy_2V2eb093jBpkEdcGBINGA1ov

-  Rocket 3.5(7759).nightly
- Device: Advan i5C
-  WebView 69.0.3497.100
-  Android: 5.1

cnevinc:
Whenever we change the app language, we change the Firebase user properties "app_lang" to the latest language. And then fetch the remote config from Firebase again ( this will also call `activateFetched()` so it'll take effect immediately )

I separate this PR with #2981 , because I think this PR may not be helpful since
1. User properties update takes time. So the test case will fail. [see below in the official document](https://firebase.google.com/docs/analytics/android/properties?hl=zh-TW#add_analytics_to_your_app) 
>Note: Once the property is registered, it can take several hours for data collected with the property to be included in reports. When the new data is available, the user property can be used as a report filter or audience definition.

2. The more value part for this PR, is that after this PR, we need to use "user property-app_lang" instead of "locale" in Firebase. Cause the real app language the users are using is "user property-app_lang" .
Something need more testing: If "user property-app_lang"  is required when user first launch the app, this user property may not be there if it'll take hours to update to Firebase. That means we can't use app_lang as an condition for first used users.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
@@ -142,6 +142,9 @@ public class SettingsFragment extends PreferenceFragment implements SharedPrefer
                 localeManager.setSelectedLocale(getActivity(), value);
             }
             TelemetryWrapper.settingsLocaleChangeEvent(key, String.valueOf(locale), TextUtils.isEmpty(value));
+            FirebaseHelper.updateAppLanguage(getActivity());
+            FirebaseHelper.refreshRemoteConfig();
+
             localeManager.updateConfiguration(getActivity(), locale);
 
             // Manually notify SettingsActivity of locale changes (in most other cases activities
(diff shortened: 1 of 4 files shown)


cnevinc:
Close this PR since we decided not to use translated strings in remote config.

PR #2990, opened 2018-12-07
Use Kinto to flip Firebase initialization. Send usage data only stop collecting the UI event.

cnevinc:
This PR may need data review. Or at least change the Term of service.

Decouple "Send Usage Data" pref with Firebase initialization.
Remote Config, Crash Report, and Push Notification are required components for the app to function.
So above features could only be stopped by Kinito.

"Send Usage Data" can now only control the UI event collecting.


cnevinc:
will not use it for now

Issue #2595, opened 2018-09-12
Send key events (e.g. tak screenshot) to Adjust for marketing campaign integration


elin-moco:
We may need to send some events to Adjust (just like Focus)
so we'll be able to integrate with some marketing campaigns and optimize towards the events.

https://sql.telemetry.mozilla.org/dashboard/firefox-rocket-indonesia-retention-rate
According to the Retention Rate Dashboard
The events we might want to send to Adjust are:
 - Take Screenshot
 - Enter Private Mode

The most recent need is the referral program planned to launch in Late October.
We might require invited users to complete certain events so that referrers can get rewarded.
This way we can ensure that users used our key features and probably increase retention.

elin-moco:
@chianger we might decide to go with adjust,
and also checked with legal, the usage is already covered in this sumo page:
https://support.mozilla.org/en-US/kb/how-do-you-use-adjust-firefox

elin-moco:
This should have higher priority then the D1 notification feature.

chianger:
We have this https://support.mozilla.org/en-US/kb/send-usage-data-firefox-mobile-devices accessible under settings> learn more . is that good enough ? @elin-moco

elin-moco:
Yes I think so.

elin-moco:
The events we planned to track on adjust are: (format: method/object/value -> Name on adjust)
 - click/toolbar/capture -> Capture Screenshot
 - change/private mode/enter -> Enter Private Mode
 - share/toolbar/bookmark -> Save Bookmkark
 - pin_shortcut/toolbark/link -> Add to Homescreen
 - change/setting/pref_default_browser {to: true} -> Set Default Browser
 - click/promote_share/share -> Share App
I've went through these with Mika, these events is covered in Category 2 “Interaction data” on the sumo page.

It would be great if we can document the events we sent in this github wiki:
https://github.com/mozilla-tw/Rocket/wiki/Telemetry#install-campaign-tracking
Also with a link to the sumo page when explaining opt-out.

elin-moco:
Add one event: FeedbackPositive

elin-moco:
Here are the Adjust event key created for SDK integration:
 - AddToHomescreen: g1tpbz
 - CaptureScreenshot: ky29gk
 - EnterPrivateMode: dccqsh
 - FeedbackPositive: uivfof
 - SaveBookmark: sj5vxj
 - SetDefaultBrowser: iyqe7a
[...]

PR #2991, opened 2018-12-07
Fix for Adjust event . See: #2595

cnevinc:
There are three commits in this PR.

The first one removes redundant source set in build.gradle.

The second one fixes the problem that Adjust didn't track any referral due to the mis-configure Adjust environment. The root cause of this bug is because I should use the value from `com.adjust.sdk.AdjustConfig.ENVIRONMENT_PRODUCTION` directly. instead of `"com.adjust.sdk.AdjustConfig.ENVIRONMENT_PRODUCTION"`

The third commit adds two missing Adjust event.


mTwTm:
Fixes #2992. @cnevinc do you have access to the sandboxed Adjust to verify there is no other problem?

cnevinc:
> Fixes #2992. @cnevinc do you have access to the sandboxed Adjust to verify there is no other problem?

I'm trying to do this with @elin-moco now

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -551,6 +551,7 @@ public void onMenuItemClicked(View v) {
                 Intent intent = new Intent(this, PrivateModeActivity.class);
                 startActivity(intent);
                 overridePendingTransition(R.anim.tab_transition_fade_in, R.anim.tab_transition_fade_out);
+                TelemetryWrapper.togglePrivateMode(true);
```
mTwTm: Consider giving this a version so we know why the volume suddenly increases.

cnevinc: added

cnevinc: I think maybe it should be a new ping. SInce it's from Tab Tray

Issue #2932, opened 2018-11-20
Disable browsingWebsiteBackAndForward_backAndFrowardToWebsite


mTwTm:
See #2855 
It's failing at a rate more than 10%, we should remove this test to stop blocking our workflow.

cnevinc:
see: https://app.bitrise.io/build/b9cebd4f9bb44e39 & https://app.bitrise.io/build/3340e111adcca247

I'll have a PR and disable it first

PR #2993, opened 2018-12-09
Remove browsingWebsiteBackAndForward_backAndFrowardToWebsite test for intermittent. see: #2932

cnevinc:
Please leave the bug open so it can be fixed in the future.
The way we check page loading complete should change from BrowserFragment's isLoading flag to SessionObserver


mTwTm:
Please check if the new output.json needed for this fix?

PR #2994, opened 2018-12-12
Create home and banner impression telemetry pings

### LINKED ISSUE #2898: Track home page visit event
**Why/User Benefit/User Problem**
We need to know how many times "home" being visited.

**User story**
As a data analyst, I would like to know how many times "home" being visited so we can know the actual conversion rate, and how it's being used so we can optimize.

1. Event shows up in Telemetry dasboard
   https://sql.telemetry.mozilla.org/queries/47388#135604

**Additional context**
Suggested event structure would be (in the format of method/object/value {extra}):
Visit homepage: show/home/tabtray|toolbar|close_all|app_open

mTwTm:
Fixes #2898 

Also, You can consider changin AppConfigWrapper.java to:
public static String getBannerRootConfig(Context context) {
-        return FirebaseHelper.getRcString(context, FirebaseHelper.BANNER_MANIFEST);
+        return "https://firebasestorage.googleapis.com/v0/b/rocket-dev01.appspot.com/o/Banner.json?alt=media&token=5ae1c5dd-fba0-4490-bf2b-08c727b5786c";
}

to test these.


app/src/main/java/org/mozilla/rocket/banner/BannerAdapter.java
```diff
@@ -100,4 +100,11 @@ public void onBindViewHolder(@NonNull BannerViewHolder holder, int position) {
     public int getItemCount() {
         return DAOs.size();
     }
+
+    public String getFirstDAOId() {
+        if (DAOs.size() <= 0) {
+            LoggerWrapper.throwOrWarn(LOG_TAG, "Invalid banner size");
```
cnevinc: do you want to return here?

mTwTm:
Per what I heard from Eddie, the purpose of the home impression is to calculate the banner conversion rate ant that's why I'm adding both.

Issue #2847, opened 2018-10-30
Redirect to website's home page in private page 


ivantw08:
**Steps To Reproduce**
1. Launch Rocket
2. Switch to Private mode
3. Visit youtube.com and play a video
4. Switch to normal mode
5. Tap Open private browsing history in the notifcation

**Expected behavior**
back to video's page

**Actual behavior**
redirect to youtube's home page 

**Screenshots/videos**
https://drive.google.com/open?id=1AJhc0DWK2miNTBuyvqOgZvobHpWLB-9i

**Device info (please complete the following information):**

- Lite 1.0.0(7916)
- Device: Redmi 4A
- WebView: 69.0.3497.100
- Android: 7.1.2

------------
**note**
Same issue happened in yahoo's website 

**Steps To Reproduce**
1. Launch Rocket
2. Switch to Private mode
3. Visit yahoo and click a news
4. Switch to normal mode
5. Tap Open private browsing history in the notification

**actual behavior**
redirect to yahoo home page

Jacklin-ux:
Expect: should back to video's page, so users no need to search the video again.

chianger:
blocker for v1.0.1

mTwTm:
Confirmed regression: Not happening on 3.0 (Private & Normal), Reproducing in 3.4.1 (Private)

mTwTm:
Move blocker tag to #2930 @cnevinc believes the resuming is a separate bug.

chianger:
this has been existing for multiple releases so it's not critical blocker. let's get it fixed in next release.

chianger:
let's block this on K2 then.

cnevinc:
This is fixed here 
https://github.com/mozilla-tw/Rocket/pull/2995

Daisy-pliu:
**Test Scenario 1 result :** ✅
[Test Step]
1. Launch Rocket
2. Switch to Private mode
3. Visit Youtube.com and play a video
4. Switch to normal mode
5. Tap Open private browsing history in the notification
[Actual behavior]
navigate to Youtube video's page

**Test Scenario 2 result :** ✅
[Test Step]
 1. Launch Rocket
 2. Switch to Private mode
 3. Visit yahoo.com and click a news
 4. Switch to normal mode
 5. Tap Open private browsing history in the notification
[Actual behavior]
navigate to Yahoo new's page

[Test Device] 
SamSung S7 Edge (8.0.0)
[Test Build]
[...]

PR #2995, opened 2018-12-13
Live data should only be responsible for the first new tab. closes #2432

### LINKED ISSUE #2432: After leaving the private mode, entering settings and tapping exit, the app returns to the private mode
Prerequisite:
There are existing tabs

STR:

1. Enter the private browsing mode
2. Tap the purple icon to leave the private mode
3. Enter settings, then tap Exit

Expected result:
The app should exit

Actual result:
The app returns to the private browsing mode

Reproduce Rate: 100%

Device: Samsung Galaxy J7 Prime
Rocket Ver.: 3.0.0(6193)
Android Ver.: 7.0
WebView: 67.0.3396.68

Clip:

[bug.zip](https://github.com/mozilla-tw/Rocket/files/2290569/bug.zip)

cnevinc:
Currently, loadUrl in PB mode is controlled by LiveData. register in onResume will trigger the url loading again. There are two approaches to fix this 
1. Update the LiveData when onPause
2. Move LiveData registration to onViewCreated. So only when Private BrowserFragment.kt is added (the user starts a new private session) will trigger loadUrl ( add a new tab, for example)

I choose approach 2. There's no reason to trigger a reload for just onResume/onPause.

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
@@ -83,11 +83,6 @@ class BrowserFragment : LocaleAwareFragment(),
             if (BuildConfig.DEBUG) {
                 throw RuntimeException("No activity to use")
             }
-        } else {
-            if (fragmentActivity is TabsSessionProvider.SessionHost) {
-                sessionManager = fragmentActivity.sessionManager
-                observer = Observer(this)
-            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


cnevinc:
closes #2847

cnevinc:
@mTwTm I thought in my PR, Observer registration did happen before LiveData binding. And I've tested that context menu did show up so I didn't see the regression. Could you please help elaborate more?  Thank you!

mTwTm:
Hmmm, what I've tried was selecting text (which was not described in the original bug) so I might be wrong.

PR #3001, opened 2018-12-14
Reduce CPU usage of ui-progress when idling. closes #2454

### LINKED ISSUE #2454: High CPU usage when idle caused by AnimatedProgressBar
See this issue:
https://github.com/mozilla-mobile/android-components/issues/622

I noticed that Rocket is affected too. Rocket does not use the `ui-progress` component yet. So a fix would need to be applied to the Rocket repository manually.

On my Nexus 9 Rocket is using about 10% - 20% of CPU while idling just because of AnimatedProgressBar.

cnevinc:
Calling DrawableWrapper.isVisible() will return true by default unless we explicitly set it to false. So when
1. Opening the app
2. Returning the app from recent apps
So the animation will start even though the progress bar is hidden.

Here I only start the animation when setting the progress and explicitly passing `visible=true, restart=true` to `DrawableWrapper.setVisible(visible, restart)`

I also remove the `onSaveInstanceState` code since
1.  It's not required under our current session. Observer architecture.
2. setProgress(0) in the old code will trigger visibility change again and start the animation unexpectedly. Which we don't expect this behaviour (setting progress by non-sessionObserver)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.java
+++ b/app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.java
@@ -233,13 +233,6 @@ public class AnimatedProgressBar extends ProgressBar {
         mIsRtl = (ViewCompat.getLayoutDirection(this) == ViewCompat.LAYOUT_DIRECTION_RTL);
     }
 
-    @Override
-    public Parcelable onSaveInstanceState() {
-        // Force update as 0 (Don't save state) for ProgressBar to avoid 0 -> 100 end animation
-        setProgress(0);
-        return super.onSaveInstanceState();
-    }
-
     private void cancelAnimations() {
         if (mPrimaryAnimator != null) {
             mPrimaryAnimator.cancel();
@@ -261,6 +254,9 @@ public class AnimatedProgressBar extends ProgressBar {
         a.recycle();
 
         setProgressDrawable(buildDrawable(getProgressDrawable(), wrap, duration, itplId));
+        if (getProgress() == 0) {
+            setVisibilityImmediately(GONE);
+        }
 
         mPrimaryAnimator = createAnimator(getMax(), mListener);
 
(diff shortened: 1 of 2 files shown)


cnevinc:
After some survey, using `restart` in `Drawable.setVisble(visible,restart)` seems to be a better option to determine if we want to really start the animation. I'll have PR later.
see : http://androidxref.com/9.0.0_r3/xref/frameworks/base/graphics/java/android/graphics/drawable/AnimatedStateListDrawable.java#88

app/src/main/java/org/mozilla/focus/widget/ShiftDrawable.java
```diff
@@ -58,10 +58,10 @@ public Animator getAnimator() {
     @Override
     public boolean setVisible(boolean visible, boolean restart) {
         final boolean result = super.setVisible(visible, restart);
-        if (isVisible()) {
+        if (visible && restart) {
             mAnimator.start();
         } else {
-            mAnimator.end();
+            mAnimator.cancel();
```
mTwTm: Wondering why we need to change this here, but I think cancel is better.

app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.java
```diff
@@ -297,6 +293,11 @@ public void onAnimationRepeat(Animator animator) {
 
     private void setVisibilityImmediately(int value) {
         super.setVisibility(value);
+        if (value == GONE) {
+            getProgressDrawable().setVisible(false, false);
+        } else {
```
mTwTm: Not sure if we want to differentiate between Visible and Invisible here.

cnevinc: we also want to stop the animation when Invisible. So let's keep it as is.

PR #3009, opened 2018-12-18
Track sideload install from partner

mxlius:


cnevinc:
Once @mTwTm's concern is solved, I'm fine with the rest. 
Current CI didn't fail because non-release builds are not using those BuildConfig  (e.g. BuildConfig.ADJUST_DEFAULT_TRACKER). But this might not be the case in the future.

PR #3020, opened 2018-12-20
Click exit in menu will also exit Private Mode session. closes #2432

### LINKED ISSUE #2432: After leaving the private mode, entering settings and tapping exit, the app returns to the private mode

cnevinc:
This PR assume sending intent with  NTENT_EXTRA_EXIT == hasPrivateSession
Maybe need to add a test to secure it.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -80,6 +80,7 @@ import org.mozilla.focus.web.WebViewProvider;
 import org.mozilla.focus.widget.FragmentListener;
 import org.mozilla.focus.widget.TabRestoreMonitor;
 import org.mozilla.rocket.component.LaunchIntentDispatcher;
+import org.mozilla.rocket.component.PrivateSessionNotificationService;
 import org.mozilla.rocket.nightmode.AdjustBrightnessDialog;
 import org.mozilla.rocket.privately.PrivateMode;
 import org.mozilla.rocket.privately.PrivateModeActivity;
@@ -99,6 +100,7 @@ import java.util.List;
 import java.util.Locale;
 
 import static android.view.WindowManager.LayoutParams.BRIGHTNESS_OVERRIDE_NONE;
+import static org.mozilla.rocket.privately.PrivateMode.INTENT_EXTRA_EXIT;
 
 public class MainActivity extends BaseActivity implements FragmentListener,
         ThemeManager.ThemeHost,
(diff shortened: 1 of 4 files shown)


cnevinc:
To achieve observable session, we can do similar things in Private Mode's BrowserFragment  like: https://github.com/mozilla-tw/Rocket/blob/b82d7e7b5838c61043ddeebb03651d24288ebac5/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java#L1286
The question is: Should we do it now, or do it later?
If we do it now, we may need to test more scenario than the original bug.
@mTwTm How do you think?

cnevinc:
After some discussion, we'll keep the PR first and adjust the structure/ session manager usage when refactorying.

app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -217,6 +218,12 @@ class PrivateModeActivity : BaseActivity(),
             finishAndRemoveTask()
             return true
         }
+        // MainActivity wants to cal onExit. We drop the browser and then show Android's home screen.
```
mTwTm: NIT: call

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -99,6 +100,7 @@
 import java.util.Locale;
 
 import static android.view.WindowManager.LayoutParams.BRIGHTNESS_OVERRIDE_NONE;
+import static org.mozilla.rocket.privately.PrivateMode.INTENT_EXTRA_EXIT;
```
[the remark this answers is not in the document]
mTwTm: ^ This comment is more suitable to be added in the overall comment rather than this review comment.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -701,7 +703,15 @@ private void onPreferenceClicked() {
 
     private void onExitClicked() {
         GeoPermissionCache.clear();
-        finish();
+        // if there's no private session, we just call finish. Otherwise, we ask PrivateModeActivity to drop
+        // BrowserFragment and show the home screen from there.
+        if (PrivateMode.hasPrivateSession(this)) {
```
mTwTm: A privateModeActivity may be created anyway even if there is no private session. This implementation does not handle that case.

mTwTm:
The approve was a misoperation. My real initiative was to comment.

mTwTm:
Closing favoring #3053

PR #3025, opened 2018-12-24
Add null check for Adjust and TelemetryWarpper

cnevinc:
applicationContext in Adjust.java may be null. And the caller for TelemetryWarpper.isTelemetryEnabled could pass null. closes #3001


app/src/release/java/org/mozilla/focus/utils/AdjustHelper.java
```diff
@@ -53,7 +56,7 @@ public static void setupAdjustIfNeeded(FocusApplication application) {
     }
 
     public static void trackEvent(String eventToken) {
-        if (TelemetryWrapper.isTelemetryEnabled(focusApplication)) {
+        if (focusApplication != null && TelemetryWrapper.isTelemetryEnabled(focusApplication)) {
```
mTwTm: Do we need this if it is already handled in isTelemtryEnabled?

Issue #3006, opened 2018-12-18
(meta) tab session refactor (dependency of #2537 multitab)


chianger:
There are multiple WIP pull requests regarding refactoring around tab sessions. 
This refactor also eases the use of android component.

chianger:
include #3008

cnevinc:
I've spent some time in this sprint for this(2-3 days).
I've looked at AC's and our(Lite)'s code base and try to narrow down the scope and make an estimation on schedule for this task.

To let us use GeckoView, there are at least three things to complete:
1. Replace our session implementation to AC's browser-session module.
2. Make WebkitView implement AC's concept-engine module, or look at how focus-android uses GeckoView
3. Hook them up in SessionManager.

To use features directly from AC, we'll need more:
4. Use AC's Session Manager and feature-session module 
5. Use AC's browser-engine-gecko to interact with engine

For (1), we had a WIP PR https://github.com/mozilla-tw/Rocket/pull/3003. though it's not completed.
For (2), I haven't deeply looked at it yet.

Step 1~3 will take 3-4 weeks to implement (including fixing all Unit Test and add Integration Test).
Step 4~5 will need longer discussion.

cnevinc:
Turns out Focus uses IWebViewProvider to provide different IWebView implementation. [see](https://github.com/cnevinc/focus-android/blob/a999b406d5971172591908a1bd9bf55f24f3afb0/app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt#L21)
And it didn't use concept-engine at all. [see](https://github.com/cnevinc/focus-android/blob/a97f4eb4d9a9ce8c3373401f8b24d27faf1df569/app/src/main/java/org/mozilla/focus/Components.kt#L28)
Question: Since TabView is similar to IWebView for the same purpose, can we use GeckoView now?

Maybe I can come out with an estimation for both situations tomorrow or later.

chianger:
this refactoring work will be mainly for #2537 so should be worked on all together.

PR #3064, opened 2019-01-11
Use SessionManager.Observer and Session.Obsever to determine if the loading is done.

cnevinc:
Related #3006
Previously we use BrowserFragment's state as the indicator. Now we should use the callback in SessionManager's
and Session's Observer.

I'll land this first so I can continue the refactoring.


app/src/androidTest/java/org/mozilla/focus/helper/PrivateSessionLoadedIdlingResource.java
```diff
@@ -4,21 +4,36 @@
...
  */
-public class PrivateSessionLoadedIdlingResource implements IdlingResource, BrowserFragment.LoadStateListener {
+public class PrivateSessionLoadedIdlingResource implements IdlingResource {
     private ResourceCallback resourceCallback;
     private boolean isCompleted;
+    private Observer observer = new Observer();
 
     public PrivateSessionLoadedIdlingResource(PrivateModeActivity activity) {
-        activity.getBrowserFragment().setIsLoadingListener(this);
+        SessionManager sessionManager = activity.getSessionManager();
```
josephyu1110: nit: final

app/src/androidTest/java/org/mozilla/focus/screengrab/PrivateBrowsingScreenshot.java
```diff
@@ -77,10 +77,12 @@ public void screenshotPrivateBrowsing() {
         // Click private home search field and enter test url
         onView(withId(R.id.pm_home_fake_input)).perform(click());
 
+        // create the idlingResource be fore the new session is created so we can register when it's created.
```
josephyu1110: typo in comment: be fore -> before

app/src/androidTest/java/org/mozilla/focus/helper/PrivateSessionLoadedIdlingResource.java
josephyu1110: nit: final

app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -119,7 +117,7 @@ class BrowserFragment : LocaleAwareFragment(),
             insets
         }
         sessionManager = TabsSessionProvider.getOrThrow( activity)
-        observer = Observer(this, loadStateListener)
```
mTwTm: Change the constructor class Observer(val fragment: BrowserFragment, private val loadStateListener: LoadStateListener) now that we don't use LoadStateListener as well, too?

app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -178,16 +176,6 @@ class BrowserFragment : LocaleAwareFragment(),
         return false
     }
 
-    fun setIsLoadingListener(listener: LoadStateListener) {
```
mTwTm: Remove the now unused LoadStateListener in BrowserFragment.java, too?

cnevinc: It's been there since day 1 (@ahunt  if I remember correctly.) I'd hesitate to remove old code unless it's required. The one added in BrowserFragment.kt is added by @josephyu1110 recently. So I'm confident we can remove it here.

For the LoadStateListener in BrowserFragement, I'll remove it when we do Session refactoring. Since this PR is for IdelingResource refinement, I'll do it separately.

PR #3073, opened 2019-01-15
add test on default browser test

Daisy-pliu:
1.  change test behavior to "set firefox lite as default when chrome is default browser"
2. add new test case "set firefox lite as default when no default browser "
3. add comment on DefaultBrowserTest

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/DefaultBrowserTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/DefaultBrowserTest.java
@@ -57,8 +57,19 @@ public class DefaultBrowserTest {
         uiDevice = UiDevice.getInstance(getInstrumentation());
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/DefaultBrowserTest.java
```diff
@@ -92,16 +104,106 @@ public void changeDefaultBrowser_prefSwitched() {
...
+        try {
+            browserList.scrollIntoView(rocketView);
+            final UiObject browserCandidate = uiDevice.findObject(rocketView);
+            browserCandidate.click();
+
+        } catch (UiObjectNotFoundException e) {
+            throw new AssertionError("Could find the Firefox Lite app in default browser", e);
+        }
+
+        // Now launch Rocket's setting activity
```
cnevinc: Firefox Lite's

PR #3074, opened 2019-01-15
Change steps in browsing history test

Daisy-pliu:
1. A combination of 3 commits.
   a. add comment and change test step on TC0050
   b. remove his_002 as it is covered in TC0050
   c. add comment on TC0050
2. add comment to TC0051
3. add comment to TC0052

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
@@ -204,7 +204,7 @@ public class BrowsingHistoryTest {
     }
 
     /**
-     * Test case no: his_004
+     * Test case no: TC0052
      * Test case name: Clear all history record
      * Steps:
      * 1. Follow steps in his_001
--- a/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
@@ -109,14 +109,19 @@ public class BrowsingHistoryTest {
     }
 
     /**
-     * Test case no: his_001
+     * Test case no: TC0050
      * Test case name: History is recorded correctly
      * Steps:
      * 1. Launch app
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
```diff
@@ -109,14 +109,15 @@ public void tearDown() {
     }
 
     /**
-     * Test case no: his_001
+     * Test case no: TC0050
      * Test case name: History is recorded correctly
      * Steps:
      * 1. Launch app
      * 2. Visit some websites
      * 3. Tap Menu
```
cnevinc: suggest: 
2: visit PATH_SITE_1
3. visit PATH_SITE_2

see the comment below.

app/src/androidTest/java/org/mozilla/focus/activity/BrowsingHistoryTest.java
```diff
@@ -154,38 +155,25 @@ public void browsingTwoWebSites_sitesAreDisplayedInOrderInHistoryPanel() {
...
-    @Test
-    public void browsingWebsiteThenTapTheFirstHistoryItem_browsingHistoryRecordCorrectly() {
-
-        final String targetUrl = browsingWebsiteAndOpenHistoryPanel();
-
-        // Click the first item in history panel, index 0 is date label so index 1 is the latest one of history item
+        // Click the first item in history panel, index 0 is date label so index 2 is the latest one of history item
         onView(ViewMatchers.withId(R.id.browsing_history_recycler_view))
-                .perform(RecyclerViewActions.actionOnItemAtPosition(1, click()));
+                .perform(RecyclerViewActions.actionOnItemAtPosition(2, click()));
```
cnevinc: I'm first confused why it's 2 here. 
After reading the code, I found that In the test case, you clicked two websites, so you want to make sure the 2nd URL is shown in the 1st position, and the 1st URL is shown in the 2nd position.  
But this behaviour is not stated in the comment of this method.

Issue #3067, opened 2019-01-14
Default search engine not set to Google in ID


cnevinc:
STR:
1. Set the default search engine to Google
2. Switching locale to ID 
3. Do a search using a keyword.

Expected:
The search engine should stay the same.

Actual
Default search engine for Google is missing. It becomes Yahoo.

Daisy-pliu:
Test Result: ✅
Test device : Redmi 4X (7.1.2)
Version: 1.0.9(9367).nightly 
Webview: 71.0.3578.99
Test step:
1. change system language to indonesia
2. launch app (clear app data)
3. go to [search engine] in [settings]
Test result:
User can see Google, Yahoo, DuckDuckGo search engines

PR #3078, opened 2019-01-15
Make sure our default search engine is Google for all locales. #3067

cnevinc:
Previously there's a bug that I forget to change the search engine code
for Google for Indonesia users. This test is to prevent it happening
again.


app/src/androidTest/java/org/mozilla/focus/activity/DefaultSearchEngineTest.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        if (defaultLocale != null) {
+            Locale.setDefault(defaultLocale);
+        }
+    }
+
+    @Test
+    public void defaultSearchEngineShouldBeGoogle() {
+
+        Locale.setDefault(new Locale(locale));
```
mTwTm: Are we going to test different locales? If we're only checking default locales then I guess we might discover this regression in manual testing as well. (We shouldn't be relying on this though.)

cnevinc:
I'll use LocaleList.BUNDLED_LOCALES instead

cnevinc:
fixed

PR #3081, opened 2019-01-16
Add telemetry pings to better understand Firebase lifecycle

cnevinc:
We want this data cause we want to verify when below timing, FIrebase RemoteConfing has completed fetching or not:
- The users see onboarding page
- The users finish onboarding page
- The users see the first level toolbar of HomeFragment-
- The users see the second level toolbar of HomeFragment
- The users see the first level toolbar of BrowserFragment.
- The users see the second level toolbar of BrowserFragment.

Confirmed with @elin-moco 
1. We only need to record the first fetch when the app is launched. 
2. We only need to record the ping for new users only (in case data mixed up)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
@@ -49,8 +49,6 @@ class FirstrunFragment : Fragment(), View.OnClickListener, Screen {
 
         exitTransition = transition
 
-        // We will send a telemetry event whenever a new firstrun page is shown. However this page
-        // listener won't fire for the initial page we are showing. So we are going to firing here.
         isTelemetryValid = true
         telemetryStartTimestamp = System.currentTimeMillis()
     }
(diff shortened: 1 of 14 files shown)


elin-moco:
Note that we only need to record the ping for new users only (in case data mixed up)

cnevinc:
@elin-moco Please help review the pings
https://github.com/mozilla-tw/FirefoxLite/pull/3081/files#diff-833b6d7070c26b0f7548f4ffdc70eb8eR3

cnevinc:
I need some tweeks in local.  Close this temporarily,

cnevinc:
> We want this data cause we want to verify when below timing, FIrebase RemoteConfing has completed fetching or not:
> 
> * The users see onboarding page
> * The users finish onboarding page
> * The users see the first level toolbar of HomeFragment
This timing is almost the same as "The users finish onboarding page"
> * The users see the second level toolbar of HomeFragment
When "The users see the first level toolbar of HomeFragment", the second level is already inflated. so the timing is the same.
> * The users see the first level toolbar of BrowserFragment.
BrowserFragment initialized earlier than HomeFragment, so this can be skipped.
> * The users see the second level toolbar of BrowserFragment.
ditto
> 
> Confirmed with @elin-moco
> 
> 1. We only need to record the first fetch when the app is launched.
> 2. We only need to record the ping for new users only (in case data mixed up)

PR #3086, opened 2019-01-21
Reduce CPU usage for AnimatedProgressbar #proposal2

### LINKED ISSUE #2454: High CPU usage when idle caused by AnimatedProgressBar

cnevinc:
closes #2454
Use Animator's callback to invalidate ProgressDrawable/ShiftDrawable

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/ShiftDrawable.java
+++ b/app/src/main/java/org/mozilla/focus/widget/ShiftDrawable.java
@@ -8,11 +8,13 @@ package org.mozilla.focus.widget;
 import android.animation.Animator;
 import android.animation.ValueAnimator;
 import android.graphics.Canvas;
+import android.graphics.Color;
 import android.graphics.Path;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


mTwTm:
Please rebase to current firefoxlite/master to get rid of a test intermittent.

app/src/main/java/org/mozilla/focus/widget/ShiftDrawable.java
```diff
@@ -43,12 +45,9 @@ public ShiftDrawable(@NonNull Drawable d, int duration, @Nullable Interpolator i
         mAnimator.addUpdateListener(new ValueAnimator.AnimatorUpdateListener() {
             @Override
             public void onAnimationUpdate(ValueAnimator animation) {
-                if (isVisible()) {
-                    invalidateSelf();
-                }
+                invalidateSelf();
```
cnevinc: maybe check callback's visibility before calling invalidateSelf?

Issue #3098, opened 2019-01-23
changeDefaultBrowser_whenNoDefault seems to fail


mTwTm:
Given this is a new test landed when Bitrise is broken, proposing to back it out until it is stable.

PR #3101, opened 2019-01-23
Remove test `changeDefaultBrowser_whenNoDefault` because of intermittent

mTwTm:
Related to #3098 but not fixing it thoroughly.


app/src/androidTest/java/org/mozilla/focus/activity/DefaultBrowserTest.java
```diff
@@ -158,7 +158,6 @@ public void changeDefaultBrowser_whenChromeIsDefault() {
      * 4. Go back to app's settings
      * 5. Check it correctly set default browser to Firefox Lite
      */
-    @Test
```
Daisy-pliu: please add @Ignore above @Test , thx

PR #3113, opened 2019-01-25
[TC0043] Turbo mode switcher in settings

Daisy-pliu:
[TC0043] Turbo mode switcher in settings


app/src/androidTest/java/org/mozilla/focus/activity/TurboModeSwitcherTest.kt
```diff
@@ -0,0 +1,39 @@
...
+     */
+    @Test
+    fun turboMode() {
+
+        setting {
+            // check initial state
+            prefTurboMode()
+            isChecked()
+            click()
+            isNotChecked()
```
cnevinc: maybe add @After to make sure you recover the app back to the state before tests.

PR #3131, opened 2019-01-30
[TC0034]menu test and [TC0049]remove file from list

Daisy-pliu:
[TC0034] menu layout test in MenuTest
[TC0049] remove a download from list in DownlaodTest


cnevinc:
per discussion, keep it as

PR #3143, opened 2019-02-11
screengrab fix and add screengrabfile

Daisy-pliu:
1. exclude screengrabfile from gitignore
2. add locale tw and vi
3. fix screengrab test because of intermittent issue

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/screengrab/BrowserFragmentScreenshot.kt
+++ b/app/src/androidTest/java/org/mozilla/focus/screengrab/BrowserFragmentScreenshot.kt
@@ -126,7 +126,7 @@ class BrowserFragmentScreenshot : BaseScreenshot() {
             loadPageFromHomeSearchField(activityTestRule.activity, TARGET_URL_LICENSE)
             longClickOnWebViewContent(activityTestRule.activity)
             clickTextActionMore()
-            checkSearchInRocketIsDisplayed()
+//            checkSearchInRocketIsDisplayed()
             takeScreenshotViaFastlane(ScreenshotNamingUtils.BROWSER_TEXT_ACTION_DAILOG)
         }
     }
--- a/app/src/androidTest/java/org/mozilla/focus/screengrab/DownloadScreenshot.java
+++ b/app/src/androidTest/java/org/mozilla/focus/screengrab/DownloadScreenshot.java
@@ -183,6 +183,9 @@ public class DownloadScreenshot extends BaseScreenshot {
                 .check(matches(isDisplayed()));
         Screengrab.screenshot(ScreenshotNamingUtils.DOWNLOAD_DOWNLOADED);
 
+        SystemClock.sleep(MockUIUtils.LONG_DELAY);
+        Screengrab.screenshot(ScreenshotNamingUtils.DOWNLOAD_DOWNLOADED);
+
         // Open menu
         // Since right now snackbar will overlap with menu bar and we don't want to wait until snackbar is dismissed,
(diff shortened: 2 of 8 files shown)


app/src/androidTest/java/org/mozilla/focus/screengrab/DownloadScreenshot.java
```diff
@@ -209,9 +212,10 @@ public void screenshotDownload() {
                 .inRoot(RootMatchers.isPlatformPopup())
                 .perform(click());
 
-        // Check if delete successfully message is displayed
-        onView(allOf(withId(android.support.design.R.id.snackbar_text), withText(containsString(IMAGE_FILE_NAME_DOWNLOADED_PREFIX))))
-                .check(matches(isDisplayed()));
+//        // Check if delete successfully message is displayed
```
cnevinc: remove unused

app/src/androidTest/java/org/mozilla/focus/screengrab/PrivateBrowsingScreenshot.java
```diff
@@ -95,7 +95,7 @@ public void screenshotPrivateBrowsing() {
         onView(withId(R.id.btn_delete)).perform(click());
 
         // Check private browsing is cleared toast is displayed
-        AndroidTestUtils.toastContainsText(activityTestRule.getActivity(), R.string.private_browsing_erase_done);
+//        AndroidTestUtils.toastContainsText(activityTestRule.getActivity(), R.string.private_browsing_erase_done);
```
cnevinc: ditto

fastlane/Screengrabfile
```diff
@@ -0,0 +1,12 @@
+
+#locales ['en','hi','in','jv','su','th','tl','zh-TW','vi']
```
cnevinc: nice!

app/src/androidTest/java/org/mozilla/focus/screengrab/DownloadScreenshot.java
Daisy-pliu: I comment it because there is intermittent that fails the test on my local from time to time.
I keep it because it probably relate to individuals CPU computing speed on desktop which may differ in different computer.

app/src/androidTest/java/org/mozilla/focus/screengrab/DownloadScreenshot.java
```diff
@@ -209,9 +212,11 @@ public void screenshotDownload() {
                 .inRoot(RootMatchers.isPlatformPopup())
                 .perform(click());
 
+        SystemClock.sleep(MockUIUtils.POPUP_DELAY);
```
cnevinc: Just wanted to note: SystemClock.sleep should only be used in Screengrab, not in regular UI test. Or we'll have too many flaky tests.

app/src/androidTest/java/org/mozilla/focus/screengrab/BrowserFragmentScreenshot.kt
```diff
@@ -126,7 +126,7 @@ class BrowserFragmentScreenshot : BaseScreenshot() {
             loadPageFromHomeSearchField(activityTestRule.activity, TARGET_URL_LICENSE)
             longClickOnWebViewContent(activityTestRule.activity)
             clickTextActionMore()
-            checkSearchInRocketIsDisplayed()
+//            checkSearchInRocketIsDisplayed()
```
cnevinc: If you really want to keep this, please state why it's important so in the future other people can remember.
But I prefer to remove it since: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

PR #3206, opened 2019-02-21
wip, test with share

cnevinc:


cnevinc:
just a sample to @Daisy-pliu , can close now

PR #3208, opened 2019-02-21
Closes #3172 - Hide Quick Search when there's no input

### LINKED ISSUE #3172: User feedback: Quick Search
Here's my first impression/feedback

1. the fact that I can tap on Facebook/Instagram/YouTube as shortcut to their website when I didn't type in anything in search field feels interesting 
2. Fennec behavior is that there has to be at least one character typed in order for the quick search to show up

cnevinc:


mTwTm:
Fixed in #3209

PR #3210, opened 2019-02-21
Closes #3202 - Handling searches with space on instagram.

### LINKED ISSUE #3202: Handling searches with space on instagram
Related to #3188, we can force search space with Instagram pages but it will not give result.

cnevinc:


mTwTm:
Fixed in #3209

PR #3090, opened 2019-01-21
ScreenshotViewer action: open web, edit, share

Daisy-pliu:
1. [TC0054] open web of a screenshot 
2. [TC0055] edit a screenshot 
3. [TC0056] share a screenshot
4. [TC0057] view info a screenshot

New function
1. Before: Insert screenshot to database
2. After: delete screenshot from database

PR #3218, opened 2019-02-25
Screenshot viewer test

Daisy-pliu:
Change branch name to run CI testing (refer original PR #3090)
[TC0054] open web of a screenshot
 [TC0055] edit a screenshot 
[TC0056] share a screenshot 
[TC0057] view info a screenshot


Daisy-pliu:
The failed reason is photo injection to database is slower than expectation, it turns out when starting the tests the testing screenshot is not yet prepared.

However if we use below `RemoveFlaky` workflow the low resolution phone will reduce system load so it inject photo successfully
-> success with  https://app.bitrise.io/build/2fa4e191dfcd8d7c

PR #3228, opened 2019-02-26
Issues/3227/fullcheck

### LINKED ISSUE #3227: Create a data source bhaskar for news portal

mTwTm:
Fixes #3227


components/utils/cachedrequestloader/src/main/java/org/mozilla/cachedrequestloader/BackgroundCachedRequestLoader.java
```diff
@@ -99,4 +99,18 @@ public void deleteCache(Context context, String subscriptionKey) {
             Log.e(TAG, "Failed to open cache directory when deleting cache.");
         }
     }
+
+    private void sleepIfTesting(Context context, boolean shouldSleep) {
```
cnevinc: now the production code will have test code. Maybe you can pass a nullable lambda here and runs it when it's not null.

components/utils/cachedrequestloader/src/androidTest/java/org/mozilla/cachedrequestloader/CachedRequestLoaderTest.java
```diff
@@ -65,7 +65,9 @@ public void testLoadAndCacheLoadFaster() throws InterruptedException {
                 latch2.countDown();
             }
         };
-        CachedRequestLoader cachedRequestLoader = new CachedRequestLoader(InstrumentationRegistry.getContext(), KEY, targetUrl, null, SOCKET_TAG, false, true);
+        Context context = InstrumentationRegistry.getContext();
+        CachedRequestLoader cachedRequestLoader = new CachedRequestLoader(context, KEY, targetUrl, null, SOCKET_TAG, false, true);
+        cachedRequestLoader.deleteCache(context, KEY);
```
cnevinc: maybe we should also recover the state of the test after the test is done.

PR #3236, opened 2019-02-27
Add telemetry when remote config completes fetching

cnevinc:
We want this data cause we want to verify when below timing, FIrebase RemoteConfing has completed fetching or not:

* The users see onboarding page
* The users finish onboarding page
* The users see the first level toolbar of HomeFragment-
* The users see the second level toolbar of HomeFragment
* The users see the first level toolbar of BrowserFragment.
* The users see the second level toolbar of BrowserFragment.

Confirmed with @elin-moco

1. We only need to record the first fetch when the app is launched.
2. We only need to record the ping for new users only (in case data mixed up)


cnevinc:
This is the PR I was working on. I previously force-push the wrong version. Now I need some tweaks to make sure it's what we want.

cnevinc:
not needed

Issue #1231, opened 2017-11-14
support landscape mode


joechengla:

mTwTm:
Just like to add that I've received an inquiry on IRC so I think I'd like to add it to our Top 20.

Jacklin-ux:
UX spec: https://drive.google.com/file/d/1sRtZs7LoBhviMhQh5SfwW5sR7KyakpkB/view?usp=sharing

chianger:
QAWanted for @Daisy-pliu to design test cases.

elin-moco:
Here is my suggested event structure (event name/method/object/value {extra})
Rotate landscape/rotate/screen/landscape {screenshots|webview}

Daisy-pliu:
[Landscape Mode Test Case](https://docs.google.com/spreadsheets/d/1AJZaeNDln6TJWjoilB3oz9AYpVkCTCkH4tXoo9oJMz8/edit?ts=5bea38ae#gid=169764628&range=C294) design is done

chianger:
Size: L considering tests.

chianger:
Per discussion this is the candidate for joseph to work on next.
Tagging sprint P, but be noted that this is P2 so non-blocking to the P release.

chianger:
Please include this website  https://games.gamezop.com/?id=Pa6YWojZa as dev testing for landscape.

elin-moco:
My suggested event naming would be: 
Enter landscape mode/change/landscape_mode/enter
Enter landscape mode/change/landscape_mode/exit {duration}
which exit may happen when:
 - back to portrait mode
 - app exit/background

Duration in milliseconds to track how much time user stays in landscape mode

josephyu1110:
@Jacklin-ux Please help to confirm this behavior: If user turn off the screen rotation from settings, and browsing webpage in landscape mode. What's the expected screen orientation?
<img width="280" alt="screen shot 2019-03-08 at 12 21 42 pm" src="https://user-images.githubusercontent.com/16300853/54007487-fef9d400-419c-11e9-9c41-b34e1f98dcf4.png">

Jacklin-ux:
Expectation: If turn off system rotation, Lite will not have landscape mode.

chianger:
carry over to sprint Q

chianger:
upgrade to P1 as this is prerequisite for game contents.

Daisy-pliu:
Verify : ✅ 
Test case TC0189 ~ TC0197
[...]

PR #3240, opened 2019-03-04
Support landscape mode.

josephyu1110:
- #1231

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -129,8 +129,7 @@
 
         <activity
             android:name="org.mozilla.focus.screenshot.ScreenshotViewerActivity"
-            android:exported="false"
-            android:screenOrientation="portrait" />
+            android:exported="false" />
 
         <!-- This will either be an actual FCM messaging service, or a dummy service that do nothing-->
         <service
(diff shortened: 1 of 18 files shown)


josephyu1110:
Using onWindowFocusChanged() in BrowserFragment to determine when to update screen orientation may not work on:
- HomeFragment is displayed
- UrlInputFragment is displayed
- FindInPage is displayed

josephyu1110:
After investigation, using a landscape BrowserFragment still need to create it in many places as current implementation does. See [comparison](https://docs.google.com/spreadsheets/d/1YtHdsKH3BAg9tiXFeLOeNbQ2JicXG0N-IWyj4a5kU_Y/edit#gid=433832423).

josephyu1110:
TBD: Need to add telemetry for private browsing as well?

cnevinc:
Per discussion with @Jacklin-ux 
We only need to care about orientation change when the user is currently looking at BrowserFragment.
This PR here also want to take care about Landscape Home-> Landscape BrowserFragment scenario. 
Maybe this could be easier.

I don't have time to review this since I'm working on Content Portal

Per discussion with @josephyu1110 , it's better if we can keep the state in another FullScreenBrowserFragment and reduce the reponsility of BrowserFragment. This need another sprint of implemtation.

PR #3243, opened 2019-03-05
unsubscribe PageSubscription when there is a Network error

### LINKED ISSUE #3242: Content portal repository doesn't handle network timeout

mTwTm:
Fixes #3242


mTwTm:
Let me think for a bit. We currently don't propagate reason through the util. If we want them it will require some big change.

cnevinc:
or just give me a callback when fail.

mTwTm:
integerStringPair.second will be empty when loading fail (integerStringPair.first == ResponseData.SOURCE_NETWORK)

PR #3247, opened 2019-03-05
Reuse HomeFragment instead of recreating it.

cnevinc:
The PR here is WIP.  Draft review is welcomed.

The main purpose of this PR is to reuse HomeFragment and BrowserFragmnet.
This gives us the benefit:
1. Keep the state of HomeFragment. So any view attached to it will remain once the user comes back from .... anywhere.
2. All views in HomeFragment can get ApplyWindowInset event like BrowserFragment. 
3. Remove popAllScreens(). Screens now need to remove itself. (DialogFragment can do this easily, URLInputFragment sill need to rely on FragmentListener)
4. ScreenNavigator give us a handy way to move between screens. It'll be great if we can use Navigation for Screen state/flow management.


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -1030,8 +1030,20 @@ public UrlInputFragment createUrlInputScreen(@Nullable String url, String parent
 
     @Override
     public HomeFragment createHomeScreen() {
```
mTwTm: Consider rename this to `showHomeScreen`

app/src/main/java/org/mozilla/focus/navigation/ScreenNavigator.java
```diff
@@ -98,7 +98,7 @@ public void onStop(@NonNull LifecycleOwner owner) {
     public void raiseBrowserScreen(boolean animate) {
         logMethod();
 
-        this.transactionHelper.popAllScreens();
+        this.transactionHelper.showBrowserScreen(animate);
```
mTwTm: You probably still want to popAllScreens:
Try: 
1. Enter home screen
2. search Google or anything.

... The InputFragment will not be dismissed.

PR #3299, opened 2019-03-13
News ui fix - #3296

### LINKED ISSUE #3296: When Close all tab it should show only home screen
**Steps To Reproduce**
1. launch app
2. swipe up to click any [news]
3. tap [tab number]
4. tap [close all tabs]

**Expected behavior**
see home page

**Actual behavior**
see home page with content news portal

 - Device: Redmi (7.1.1), Pixel 2.0 (9.0)

**Test version**
1.1.4 (10676).nightly 

**Reproduce rate**
10/10

cnevinc:


app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -216,6 +211,15 @@ class ContentPortalView : CoordinatorLayout, ContentAdapter.ContentPanelListener
         }
     }
 
+    fun onResume() {
```
mTwTm: Not sure if intended: But clicking home and return will close the panel

Issue #3070, opened 2019-01-15
Can not see image downloaded snack bar in private mode


Daisy-pliu:
**Steps To Reproduce**
1. launch app 
2. open private mode
3. long click any image 
3. click [save image]

**Expected behavior**
ask for access permission, show snack bar [file donwloaded] and [open]

**Actual behavior**
not show  snack bar, only notification shows downloaded image completed

**Device info (please complete the following information):**
 - Pixel 2 (9.0)
 - Redmi 4X (7.1.2)
 - Vivo 1606 (6.0.1)

**Reproduce rate**
10/10

chianger:
see also #2519 and #2645

chianger:
it's actually a feature not implemented in private mode. 
tagging P1 as we want to improve Private mode for our users.

chianger:
QAWanted for test case

chianger:
looks like #2653 is relevant

josephyu1110:
Will add setDestinationInExternalPublicDir to download request in private session to let downloaded file is openable.

mTwTm:
Push back to the next sprint due to conflict. (It's also still under review.)

Daisy-pliu:
**verify:**❌
**Test version**
RC 1.1.4 (10373)
**Test step :** 
1. launch app
2. open https://mozilla.slack.com/archives/DE18AFX53/p1549184862000200 in private mode 
3. long click any image 
4. click [save image]

**Expected behavior:**
ask for access permission, show snack bar ,show notification [file donwloaded] and [open]

**Actual behavior**
not show snack bar,show access permission, no notification to show image downloaded  - Google Pixel 2 (9.0)
not show snack bar, show access permission, only show notification to show image downloaded  - Redmi 4x (7.1.2)

Reproduce rate
5/5

Daisy-pliu:
**Verify:** ❌
**Test version**
RC 1.1.4 (10373)
**Test step :** 
1. launch app
2. open https://downloadfiletest-eab8a.firebaseapp.com/ in private mode 
3. long click any image 
4. click [save image]

**Expected behavior:**
ask for access permission, show snack bar ,show notification [file donwloaded] and [open]

**Actual behavior**
not show snack bar, show ask for access permission, only show notification for image downloaded  - Redmi 4x (7.1.2)

**Reproduce rate**
5/5

Daisy-pliu:
[...]

PR #3301, opened 2019-03-13
Fix private mode cannot get notified by relocate finish intent. Close…

josephyu1110:
…s #3070

- Private mode is in another process use context to sendBroadcast instead of local broadcast.


cnevinc:
we shouldn't expose the broadcast receiver and should only use a local broadcast receiver for both PrivateModeActivity and MainActivity.

Suggest using a factor to generate a receiver for each activity to use.

josephyu1110:
DownloadCompleteReceiver and RelocateService are running in app process, but PrivateModeActiivty is running in private_mode process which cannot receive local broadcast from app process. Under current architecture, no good way to pass PrivateModeActivity's context into DownloadCompleteReceiver or RelocateService.

PR #3311, opened 2019-03-13
bump version to 1.2

chianger:
bump version to 1.2


chianger:
fix #3310

mTwTm:
Let's use 1.2.0 instead

chianger:
fixed #3310

chianger:
fixes #3310

PR #3328, opened 2019-03-15
Enable Private Tabs in Dev Build

cnevinc:
Around the PR, I've added
1. A shim around Private Tabs so it can be remote and local configurable.
2. I reuse the current layout file cause I don't want to duplicate a lot of xml files.
3. tested with Content Portal, everything looks good.
4. No major bug found so I think it's good for landing (in Dev build only and ship to production)


cnevinc:
After internal discussion with @mxlius, we want to put the PR in separate Fragments and Activity.
But @walkingice mentioned that this lost meaning for this refactoring. Since the next step is adopting Android-components' session feature, we will still stock on the implementation in Private Mode and can't move anything further. Since @pharspher will continue the work with session refactoring, I'll leave the decision to him and stop the tab-session refactoring work here.

PR #3360, opened 2019-03-21
Closes #3179 - Need to be able to track preset topsites clicks 

### LINKED ISSUE #3179: Need to be able to track preset topsites clicks in telemetry
**User story**
As a PM (I'm writing on behave of Joe), I would like to precisely track partner link clicks and preset top site icon clicks () so that we can cross compare with partners' report and verify the integration, and make sure we are getting real conversions/revenue without being affected by integration issues.

1. Verified partner link click event ping on Firebase debugging console.
2. Be able to create reports/dashboards with the event ping on data tools we have (Amplitude/Redash)
3. Be able to cross compare with reports from partners

cnevinc:
Uses a field "isDefault" for default topsite.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/history/model/Site.java
+++ b/app/src/main/java/org/mozilla/focus/history/model/Site.java
@@ -36,6 +36,10 @@ public class Site {
     @NonNull
     private String url;
 
+    @Ignore
+    @NonNull
+    private boolean isDefault = false;
+
     @ColumnInfo(name = "view_count")
     private long viewCount;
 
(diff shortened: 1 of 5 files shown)


cnevinc:
I want to double check about below before going to review
1. What will happen if the users remove all top site and load the default top sites again? I assume we'll keep collecting them when they are clicked.
2. What will happen if the users' browsing history happens to be the same as the preset top site? I guess we shouldn't collect those?

I think current PR is okay but let's wait and review later.

cnevinc:
per discussion with @pharspher , I'll send the PR as is and land it before @pharspher's pinned top site implementation,

PR #3361, opened 2019-03-21
Pin site & News title strings

Jacklin-ux:


app/src/main/res/values/strings.xml
```diff
@@ -509,6 +516,7 @@
     <!-- Content portal -->
     <!-- Title of news portal page -->
     <string name="label_menu_news">News</string>
+    <string name="label_menu_latest_&_top_trending_news">Latest & Top Trending News</string>
```
benjamin-cheng: Please remove the `&` from string id and use escaped character `&amp;` to replace `&` in the string content.

```<string name="label_menu_latest_top_trending_news">Latest &amp; Top Trending News</string>```

PR #3367, opened 2019-03-22
Firebase refactor/fullcheck

cnevinc:
Related #3349
Not ready for merge. But a draft review is welcomed.

PR #3375, opened 2019-03-26
Separate "Send Usage Data" with "Firebase Cloud Messaging" and "Firebase Remote Config"

cnevinc:
will replace #3367


cnevinc:
while waiting for review, I"ll add more fix to simplify the API

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -78,223 +76,115 @@
...
-    public static void init(final Context context, boolean enabled) {
+    public static void init(final Context context, boolean enabled, FirebaseContract contract) {
 
-        if (getInstance() == null) {
-            initInternal(new FirebaseHelper());
+        if (firebaseContract == null) {
+            firebaseContract = contract;
         }
-        bind(context, enabled);
+        firebaseContract.setRemoteConfigDefault(contract.getRemoteConfigDefault());
```
pharspher: From the definition of the FirebaseContract, any of its instances will have their own default config during construction. 
Based on the above statement, in the case when firebaseContract is null, the call to setRemoteConfigDefault seems redundant.
On the other hand, when firebaseContract is not null, it's possible the new contract and the previous one is of different implementations (the logic here allows this possibility), overriding the original config with config from another contract seems weird to me.

cnevinc: fixed

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -79,223 +70,151 @@
...
+                final String pref = applicationContext.getString(R.string.pref_s_news);
+                final String source = firebaseContract.getRcString(pref);
+                final Settings settings = Settings.getInstance(applicationContext);
+                final boolean canOverride = settings.canOverride(PREF_INT_NEWS_PRIORITY, Settings.PRIORITY_FIREBASE);
+                Log.d(NewsSourceManager.TAG, "Remote Config fetched");
+                if (!TextUtils.isEmpty(source) && (canOverride || TextUtils.isEmpty(settings.getNewsSource()))) {
+                    Log.d(NewsSourceManager.TAG, "Remote Config is used:" + source);
+                    settings.setPriority(PREF_INT_NEWS_PRIORITY, Settings.PRIORITY_FIREBASE);
+                    settings.setNewsSource(source);
+                    NewsSourceManager.getInstance().setNewsSource(source);
```
pharspher: Just to leave a note, firebaseContract.enableRemoteConfig() only calls back if the config is fetched successfully. In case the operation is failed, the block here will not be executed, i.e. setNewSource() here will not be called. Since the method getNewsSource() will block until setNewsSource() is called, in another word, it's possible the caller of getNewsSource() will be blocked if the remote config is not fetched successfully.

cnevinc: I'll fix it in https://github.com/mozilla-tw/FirefoxLite/issues/3472

Issue #3264, opened 2019-03-07
Fix test "saveImageThenDelete_imageSaveAndDeleteSuccessfully"


Daisy-pliu:
**Describe the bug**
On Nexus 5X (8.0) 

**Steps To Reproduce**
Test case no: TC0187
Test case name: Save image
Steps:
1. Launch Rocket and visit a website with images
2. Long press on the selected image
3. Tap on "Save image"
5. Click menu
6. Click download manager
7. Check download item matches as expected

**Expected behavior**
it can see the loaded image in download manager

**Actual behavior**
when open download manager, it does not see any image downloaded 

**BItrise Screenshots/videos**
[video](https://app.bitrise.io/build/7974d8cba46cc6ec )  
failing time in video: 10:53

**Device info**
- Emulator Nexus 5X API 26 (8.0)

**Reproduce rate**
3/10

cnevinc:
not working on it so un-assign myself

PR #3384, opened 2019-03-27
Issue #3264 - Mark flaky ui test before we have idling resource 

cnevinc:


Daisy-pliu:
LGTM

Daisy-pliu:
the build failed will merge when it is fixed

PR #3386, opened 2019-03-28
Remote configurable Life feed URL

### LINKED ISSUE #3350: Remote configurable Life feed URL
**User story**
As life feed service provider, I would like to be able to use Firebase as remote config backend to configure service URL (API)

** acceptance criteria **
1. Supports versioning of the API URL to map to corresponding client app

benjamin-cheng:
Close #3350

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
+++ b/app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
@@ -7,10 +7,15 @@ package org.mozilla.focus.utils;
 
 import android.content.Context;
 
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
 public class AppConfigWrapper {
     static final int SURVEY_NOTIFICATION_POST_THRESHOLD = 3;
     static final boolean PRIVATE_MODE_ENABLED_DEFAULT = true;
     static final boolean LIFE_FEED_ENABLED_DEFAULT = false;
+    static final String LIFE_FEED_PROVIDERS_DEFAULT = "";
 
     /* Disabled since v1.0.4, keep related code in case we want to enable it again in the future */
     private static final boolean SURVEY_NOTIFICATION_ENABLED = false;
@@ -99,6 +104,26 @@ public class AppConfigWrapper {
 
     public static boolean isLifeFeedEnabled(Context context) {
         return FirebaseHelper.getRcBoolean(context, FirebaseHelper.ENABLE_LIFE_FEED);
+    }
+
+    public static String getLifeFeedProviderUrl(Context context, String provider) {
+        String source = FirebaseHelper.getRcString(context, FirebaseHelper.LIFE_FEED_PROVIDERS);
+        String url = "";
+
+        try {
+            JSONArray rows = new JSONArray(source);
+            for (int i = 0; i < rows.length(); i++) {
+                JSONObject row = rows.getJSONObject(i);
+                if (row.getString("name").equalsIgnoreCase(provider)) {
+                    url = row.getString("url");
+                    break;
+                }
+            }
+        } catch (JSONException e) {
+            e.printStackTrace();
+        }
+
+        return url;
     }
 
     static String getShareAppDialogTitle(Context context) {
--- a/app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
+++ b/app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
@@ -65,6 +65,7 @@ final public class FirebaseHelper extends FirebaseWrapper {
     static final String FIRST_LAUNCH_TIMER_MINUTES = "first_launch_timer_minutes";
     static final String FIRST_LAUNCH_NOTIFICATION_MESSAGE = "first_launch_notification_message";
     static final String ENABLE_LIFE_FEED = "enable_life_feed";
+    static final String LIFE_FEED_PROVIDERS = "life_feed_providers";
 
     private static final String FIREBASE_WEB_ID = "default_web_client_id";
     private static final String FIREBASE_DB_URL = "firebase_database_url";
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
```diff
@@ -101,6 +106,26 @@ public static boolean isLifeFeedEnabled(Context context) {
...
+            JSONArray rows = new JSONArray(source);
+            for (int i = 0; i < rows.length(); i++) {
+                JSONObject row = rows.getJSONObject(i);
+                if (row.getString("name").equalsIgnoreCase(provider)) {
+                    url = row.getString("url");
+                    break;
+                }
+            }
+        } catch (JSONException e) {
+            e.printStackTrace();
```
cnevinc: I think you should return the default value in this case

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -219,6 +220,9 @@ protected Void doInBackground(Void... voids) {
                         settings.setNewsSource(source);
                         NewsSourceManager.getInstance().setNewsSource(source);
                     }
+
+                    final String url = AppConfigWrapper.getLifeFeedProviderUrl(applicationContext, settings.getNewsSource());
+                    ThreadUtils.postToMainThread(() -> NewsSourceManager.getInstance().setNewsSourceUrl(url));
```
cnevinc: why post to main thread?

app/src/main/java/org/mozilla/rocket/content/NewsSourceManager.java
```diff
@@ -67,6 +73,15 @@ public void setNewsSource(String newsSource) {
         ContentRepository.reset();
     }
 
+    public String getNewsSourceUrl() {
+        return newsSourceUrl;
+    }
+
+    public void setNewsSourceUrl(String newsSourceUrl) {
+        this.newsSourceUrl = newsSourceUrl;
+        ContentRepository.resetSubscriptionUrl(newsSourceUrl);
```
cnevinc: why not reuse reset()?

components/partners/partnerrepository/src/main/java/org/mozilla/lite/partner/Repository.java
```diff
@@ -50,17 +52,23 @@ public void loadMore() {
         nextSubscription();
     }
 
-    // Not used yet and not tested yet.
     public void reloadData() {
         itemPojoList = new ArrayList<>();
         cacheIsDirty = true;
+        currentPage = firstPage;
```
cnevinc: can you call reset and then ` nextSubscription();` ?

app/src/main/java/org/mozilla/rocket/content/ContentRepository.kt
```diff
@@ -32,14 +31,19 @@ class ContentRepository {
...
+            INSTANCE?.setSubscriptionUrl(subscriptionUrl)
+        }
+
         @JvmStatic
         fun isEmpty() = INSTANCE == null
 
         private fun buildRepository(context: Context): Repository<out NewsItem> {
             return if (NewsSourceManager.getInstance().newsSource == NEWS_DB) {
-                RepositoryBhaskar(context)
+                RepositoryBhaskar(context, NewsSourceManager.getInstance().newsSourceUrl)
```
cnevinc: Can we derive the URL from the current source and it's URL in AppConstant? I'll like to remove the NewsSourceManager in the future cause it lacks testing and make the URL/Source hard to sync.

app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
benjamin-cheng: Empty url may be better than broken(default) url

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
benjamin-cheng: setNewsSourceUrl() will trigger series actions related to UI update.
setNewsSourceUrl() -> Repository.nextSubscription() -> onDataChangedListener::onDataChanged() -> ContentViewModel.item.setValue()

app/src/main/java/org/mozilla/rocket/content/NewsSourceManager.java
benjamin-cheng: For this case, it should not set `onDataChangedListener` as null in reset()

components/partners/partnerrepository/src/main/java/org/mozilla/lite/partner/Repository.java
benjamin-cheng: For this case, it should not set onDataChangedListener as null in reset()

app/src/main/java/org/mozilla/rocket/content/ContentRepository.kt
benjamin-cheng: Maybe we could refine the class responsibility when we try to remove NewsSourceManager. Currently, it should be reasonable to keep url in NewsSourceManager.

cnevinc:
Could you rename the commit to align below style?
Closes #3350 - Your commit message.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
cnevinc: Please correct me if I'm wrong. 
Repository.nextSubscription() -> Could run in background thread
onDataChangedListener::onDataChanged() -> it's called in LiveData's onChanged so it's UI thread.
ContentViewModel.item.setValue() -> same as above

benjamin-cheng:
OK, rename and squash the commits, thanks for the review 💯

PR #3401, opened 2019-04-01
Wrap sessionStart/End to fix IllegalStateException

mxlius:
The root cause is in Mozilla Telemetry library, exception throws when sessionStart/End overlapped.

The idea to throw exception is correct. but the API usage should contain an parameter Object as key to record the ownership of start/end. Otherwise, library user may have to wrap it up themselves just like this patch does. The cons of this work-around is that total session duration could be extended to next session(Activity resume/pause), and we won't have this problem when the library support Object key.


chianger:
Tagging sprint R for this is top crash.

PR #3405, opened 2019-04-01
Description of service provided by which partner

Jacklin-ux:


mTwTm:
If it's me I would probably try `Service provided by %1$s` because not every language put the partner name at the very end. @Jacklin-ux @cnevinc 
Also, it would be great if the spec or some descritpion could be provided so I can check if my change is valid next time. (I'm imaging it could be possible that we want to put a logo instead of phrases, then it would be a different story.)

mTwTm:
I'll keep the translation as is assuming it's non-changable. Let's change the copy if it's actually flexible.

Jacklin-ux:
In current design, we use partner's logo at the end and only provide Indonesian language. FYI

Issue #3181, opened 2019-02-18
Track organic users in telemetry


elin-moco:
**User story**
As a PM (I'm writing on behave of Joe), I would like to track product usage of organic users separately, so the usage trend will less likely to be affected by acquisition campaigns/activities, and we will be able to better track the progress of product improvements by metrics change over time/build.

**Acceptance Criteria (how do I know when I’m done?)** 
1. Verified attribute representing organic(or other channel) users in user properties on Firebase debugging console.
2. Be able to create reports/dashboards with the user properties using the data tools we have (Amplitude/Redash).
3. Be able to track product improvements by metrics change over time/build.

cnevinc:
I only send adjust tracker token as part of the important settings.
We didn't create another field in core ping cause changing the schema is tedious.

for 
> Verified attribute representing organic(or other channel) users in user properties on Firebase debugging console.

It's not in the PR cause we'll send Adjust tracker token to Firebase.  Are we allowed to do this?

chianger:
this has merged by #3406

Daisy-pliu:
As we would like to verify this feature in advance, a release APK built is handover to data engineer to make final verification.
Furthermore, on release day @Conjuror will take the effort to make sure the feature is good to ship.

PR #3406, opened 2019-04-01
Issue #3181 - Track organic users in telemetry.

cnevinc:
Send adjust tracker token as part of the important setting.
We didn't create another field in core ping cause changing the schema is tedious.

Confirmed with @elin-moco 
We query the tracker token at our best effort for simplicity. If we want to filter out scenarios like re-attribution or AttributionChanged, we'll need to do it from the query.


app/src/release/java/org/mozilla/focus/utils/AdjustHelper.java
```diff
@@ -7,22 +7,23 @@
...
 import org.mozilla.focus.BuildConfig;
 import org.mozilla.focus.FocusApplication;
+import org.mozilla.focus.R;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 
 import javax.annotation.Nullable;
 
 public class AdjustHelper {
 
+    private static final String UNEXPECTED_TOKEN = "unexpected";
```
benjamin-cheng: Defined but no usage.

PR #3412, opened 2019-04-03
Content Portal - E-Commerce

### LINKED ISSUE #3380: Simple e-commerce portal in life feed
This is a simple version given #1999 is to big. 
Story: As ID users, I would like to have the category URLs provided by e-commerce site.
AC1: onboarding is given, per ID MNC.

cnevinc:
Closes #3380


elin-moco:
The suggested event structure would be:
  - Open lifefeed EC: open/panel/lifefeed_ec
  - Click on EC item: click/panel/lifefeed_ec {pos, source={bukalapak|tokopedia|...}, category={pulsa|data|game|train|flight|event|...}}

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -42,6 +44,7 @@ class ContentPortalView : CoordinatorLayout, ContentAdapter.ContentPanelListener
 
     sealed class ContentType {
         object News : ContentType()
+        object Ticket : ContentType()
```
benjamin-cheng: Just wonder the idea behind the scene to call it `Ticket`? Would it be better to call it `Product` as a more general item?

app/src/main/java/org/mozilla/focus/utils/NewFeatureNotice.java
```diff
@@ -112,4 +118,11 @@ private void setLastShownFeatureVersion(int featureVersion) {
     public int getLastShownFeatureVersion() {
         return preferences.getInt(PREF_KEY_INT_FEATURE_UPGRADE_VERSION, 0);
     }
+
+    /**
+     * @return true if we have the E-Commerce Ticket feature and haven't shown the on-boarding
+     */
+    public boolean shouldShownEcTicketOnboarding() {
```
benjamin-cheng: Shown -> Show
`shouldShowEcTicketOnboarding`

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
cnevinc: This version is for e-voucher and ticket.
Basically, they are links to the e-commerce site, which contain tickets, vouchers, and maybe more in the future. I call them tickets just for easy identify "The links under voucher & ticket"

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -66,6 +66,7 @@
     static final String FIRST_LAUNCH_NOTIFICATION_MESSAGE = "first_launch_notification_message";
     static final String ENABLE_LIFE_FEED = "enable_life_feed";
     static final String LIFE_FEED_PROVIDERS = "life_feed_providers";
+    static final String STR_E_COMMERCE_TICKETS = "str_e_commerce_tickets";
```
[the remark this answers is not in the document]
cnevinc: I think we should have a new convention. All remote config and shared preference should have data type in its name. So we can call the right getter from its name

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
cnevinc: Changed to Shopping Link

Issue #3454, opened 2019-04-12
on-boarding shouldn't block pendingUrl


cnevinc:
this is a follow up for https://github.com/mozilla-tw/FirefoxLite/pull/3453#issue-269902695

Previously, pendingUrl is used for:
1. In `onCreate()` cache the Url from the coming intent and display on-boarding if needed
2. In `onNewIntent()`, cache the Url from the coming intent and display on-boarding if needed
3. In `onNewIntent()`, cache the Url from the coming intent load the Url after fragments are resumed.
#3452 removed scenario 2.
Now, this PR removed scenario 1.
Now pendingUrl is only used for scenario 3.

Since Firstrun could come late and the user had already started to use the app. We should load the Url directly when the intent comes and don't wait for on-boarding is shown.

ps @Jacklin-ux has confirmed this behaviour change.

Daisy-pliu:
@Jacklin-ux  could you double check the test steps? 

**Test Case**
1. vpn in indonesia
2. launch app 1.5.1
3. swipe up to see life feed
4. kill app
5. launch and see [discover new things] on boarding
6. open gamail app
7. long click on any text
8. tap [web search] -> [firefox lite]

**Expect Result:**
should see requested text search page on Lite successfully

benjamin-cheng:
@cnevinc would you like to merge PR #3455 to close this issue?

Daisy-pliu:
@Jacklin-ux  found app hanging at step 8 when perform [test case ](https://github.com/mozilla-tw/FirefoxLite/issues/3454#issuecomment-489917137) on Piexl3, Android 9, Chrome (74.0.3729.136), Lite 1.5.1 (11850). 

However I can not reproduce his issue on my test device for Samsung 8.0, Pixel 2 (9.0), on Version 1.5.1 (11850) nightly, webview: 74.0.3729.92.

cnevinc:
> However, if user do more steps as below
> 9. tap [recent app]
> 10. tap [firefox lite nightly] app -> a

Did you test the local build? Cause I think #3455 is not merged yet.

Daisy-pliu:
Since some code were landed I was doing a pretesting to have some insight and confirm test steps. Additionally some was saying it is finished and merged long time ago which is probably a miscommunication. 
In general the testing will be conducted by the time the ticket is closed.

Daisy-pliu:
verify : ✅
[...]

PR #3455, opened 2019-04-12
Issue #3454 - on-boarding shouldn't block pendingUrl

cnevinc:
Please give me some early review before I start to write unit tests for this change.

tested :
Launch Lite from Launcher / Web-search / open link in the Messenger App when:
An onboarding is displaying / the app is not started.
Maybe we can add unit tests to make sure this behaviour won't change.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -113,6 +113,7 @@ public class MainActivity extends BaseActivity implements FragmentListener,
 
     private PromotionModel promotionModel;
 
+    // Url request from onNewIntent() need to wait till fragments are resumed.
     private String pendingUrl;
 
     private BottomSheetDialog menu;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -188,14 +189,9 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-                if (Settings.getInstance(this).shouldShowFirstrun()) {
-                    pendingUrl = url;
-                    this.screenNavigator.addFirstRunScreen();
-                } else {
-                    boolean openInNewTab = intent.getBooleanExtra(IntentUtils.EXTRA_OPEN_NEW_TAB,
-                            false);
-                    this.screenNavigator.showBrowserScreen(url, openInNewTab, true);
-                }
+                boolean openInNewTab = intent.getBooleanExtra(IntentUtils.EXTRA_OPEN_NEW_TAB,
```
mTwTm: this.screenNavigator.addFirstRunScreen(); is missed for this case?

mxlius: I think this is intended behavior change, details in #3454.
When the launch intent is VIEW, load and show the URL have higher priority, so here removed the addFirstRunScreen part and also pendingUrl cache.

Issue #3456, opened 2019-04-12
R8 and byte-level optimizations should be enabled for release builds


colintheshots:
**User story**
As a user, I would like the released app to run faster due to optimizations. All Mozilla Android products should use R8 and byte-level ProGuard optimizations.

**Acceptance Criteria (how do I know when I’m done?)** 
Both R8 and byte-level optimizations are enabled in release builds.

**Additional context**
Switching to R8 will significantly improve release build times. https://android-developers.googleblog.com/2018/11/r8-new-code-shrinker-from-google-is.html

Enabling byte-level optimizations will improve application run times.

mTwTm:
With #3457 we've seen that turning R8 directly on gives us a size-regression. We might want to take a look at if we can avoid that hit which will happen soon when Android Gradle Plugin 3.4 is out.

mTwTm:
and AGP is now live on stable 😅

mTwTm:
Resolved in AGP 3.4

mTwTm:
Duplicate to #3502

mTwTm:
Oh actually the byte-level optimizations part is not a duplicate and it will give us a big size shrink! @colintheshots can you send another pr with just the `'proguard-android-optimize.txt'` part once #3503 is resolved?

chianger:
Not sure if this is still a valid ticket and requires efforts?

mTwTm:
It has limited engineering efforts but we might want a thorough test to make sure nothing breaks. It's a ~0.2 MB size improvement if we think it's worth it.

PR #3457, opened 2019-04-12
Issue #3456: R8 and byte-level optimizations should be enabled

colintheshots:


mTwTm:
This pr is failing at check apk size (3901569 > 3879731.2)

mTwTm:
Actually pretty weird, I'm not expecting turning R8 on to have size regression but I don't have time for a deep dive into apk at this moment.

mTwTm:
The last successful master build says: 3799342 so a 0.1MB regression I guess

colintheshots:
That's definitely strange. I think Google's goal for R8 was smaller APK size and a faster build. Worst case, one could take the ProGuard optimizations without R8. But R8 will become default when Android Studio 3.4 comes out.

mTwTm:
Yeah, I've seen that and I think Gradle Plugin 3.4 is already at beta stage so we will take that hit very soon anyway. Thanks for help setting up this so we have some more time to examine! I'll send this to feature grooming to decide when we can tackle this.

colintheshots:
Have you tried R8 full mode? It may take more QA time, but getting that working would potentially strip more than ProGuard does.

mTwTm:
Hmmm, I'll try to take a look at if we're running into that exact issue. If so, we might want to postpone the upgrade of AGP 3.4 or disable R8 atm.

mTwTm:
I have not deep dive into this yet but seems like upgrading to AGP 3.4 will resolve the regression
Screen Shot 2019-04-23 at 11 33 10 AM

mTwTm:
Still don't know why but seems like the `classes.dex` is bloated and the sole root cause.
Screen Shot 2019-04-23 at 11 53 05 AM
The unzipped file size of classes.dex in these two apks are 4.9 and 5.2, accordingly so this is kinda expected.
All dexes are bloating without specific tendancy flavoring specific files.
Screen Shot 2019-04-23 at 12 00 40 PM
Screen Shot 2019-04-23 at 12 00 52 PM
So, I'll basically conclude that this is based on some AGP 3.2 bug 😆

mTwTm:
favoring #3503

colintheshots:
It's not exactly a bug in AGP 3.2. Upgrading AGP is how you get new versions of R8 with better size savings and optimizations. AGP 3.5 has a huge performance fix we want on Fenix for Kotlin Coroutines, but I don't want to use an alpha version.

mTwTm:
Yeah, I'm viewing it as a bug since it's not optimizing at least par with proguard but the team also made it clear that r8 is experimental before 3.4 so my bad. We will first start at 3.4 instead of 3.5 for the same reason since it is improving already.

PR #3479, opened 2019-04-17
Issue #2834 - Enable tracking protection by default during private browsing

### LINKED ISSUE #2834: Force tracking protection when in Private Mode
Force tracking protection in PB even if the user turned off Turbo Mode in normal browsing to enhance privacy in PB

pharspher:
Close #2834

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/preference/Preferences.kt
+++ b/app/src/main/java/org/mozilla/rocket/preference/Preferences.kt
@@ -1,0 +1,67 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.rocket.preference
+
+import android.content.SharedPreferences
+
+interface Preferences {
+    fun putString(key: String, value: String)
+    fun getString(key: String, defaultValue: String): String?
+
+    fun putInt(key: String, value: Int)
+    fun getInt(key: String, defaultValue: Int): Int
+
+    fun putLong(key: String, value: Long)
+    fun getLong(key: String, defaultValue: Long): Long
+
+    fun putFloat(key: String, value: Float)
+    fun getFloat(key: String, defaultValue: Float): Float
+
+    fun putBoolean(key: String, value: Boolean)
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/preference_keys.xml
```diff
@@ -75,4 +75,7 @@
 
     <!-- Unread Myshot -->
     <string name="pref_has_unread_my_shot" translatable="false"><xliff:g id="preference_key">pref_has_unread_my_shot</xliff:g></string>
+
+    <!-- Private Mode -->
+    <string name="pref_key_pb_turbo_enabled" translatable="false"><xliff:g id="preference_key">pref_key_pb_turbo_enabled</xliff:g></string>
```
cnevinc: suggest adding data type in pref-key-name

app/src/main/java/org/mozilla/rocket/preference/Preferences.kt
```diff
@@ -0,0 +1,67 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.rocket.preference
+
+import android.content.SharedPreferences
+
+interface Preferences {
```
cnevinc: I like this interface cause we can do unit testing on shared preference.
Just one thing:
Should we expose getString(key, defaultValue) api? or just restrict the caller to use something like `shouldShowOnboarding() : Boolean`

the later give us better control on how callers can use this new Preferences interface.

app/src/main/res/values/preference_keys.xml
```diff
@@ -76,4 +76,7 @@
 
     <!-- Unread Myshot -->
     <string name="pref_has_unread_my_shot" translatable="false"><xliff:g id="preference_key">pref_has_unread_my_shot</xliff:g></string>
+
+    <!-- Private Mode -->
```
cnevinc: Private Mode 
-->
Turbo mode setting in Private Mode 
?

app/src/main/java/org/mozilla/rocket/preference/Preferences.kt
pharspher: Preferences is more like a low-level interface with get/put abstraction, so that we can have the 
 ability to change the underlying implementation at will. PrivateBrowsingSettings and GlobalSettings, by contrast, provide more high-level methods and also restrict the access to Preferences.

Issue #1999, opened 2018-05-25
Experiment: Test product promotion from e-commerce partners as part of Content Portal


joechengla:
**User story**
As Mozilla, we'd like to offer e-commerce contents to users as part of the content portal offering

**Acceptance Criteria (how do I know when I’m done?)**
1. On the UI level, users can access the content portal and e-commerce offerings will be seen
2. We have properly implemented UI event telemetry
3. We have properly implemented partner referral code, if applicable

Jacklin-ux:
Content portal concept "Subscribe your favorite categories" : https://drive.google.com/drive/folders/1yEHc8-KV1iwx8VXHNYELUViJml4r6Sae?usp=sharing

chianger:
QAWanted for @Daisy-pliu to design test cases.

Daisy-pliu:
checked with @Jacklin-ux that the detail flow is not yet finalized, QA will wait until the UX design is locked is done.

elin-moco:
@joechengla Do we want to track user-defined interests here?
Or just suggested ones for now?

chianger:
not yet able to design cases given current limited spec.

Jacklin-ux:
ux spec: https://drive.google.com/file/d/1Ik6lQDLHOrJKcbF-NCphcVUVO6hcaBm3/view?usp=sharing

(WIP)
Need to pick up what coupons/promo we would like to show in Life feed?
 
Some questions want to know:
1. Who is our target users?
2. What kind of coupons/promo they like? 
3. Do we design "sort by category" feature? or just show by due date?
...

TLHuang:
The visual spec is ready, link https://drive.google.com/drive/folders/1uVpAp7RC3Je9lZ5-Zo2dYKnCjsbV_TrX?usp=sharing

chianger:
Changed to Size: wanted since we split some efforts to a separate spike #3229

Jacklin-ux:
UX spec of "E-Voucher & Tiket": https://drive.google.com/file/d/1qGuGcZK4VM6qD8klaf-xfMFLeLvQhbT3/view?usp=sharing

elin-moco:
From marketing perspective, we may need this ping to help answer which kind of promotion have better performance (CTR/revenue).

Per discussion with Nevin, the suggested event structure would be:
 - Open lifefeed promo: open/panel/lifefeed_promo
[...]

PR #3510, opened 2019-04-23
Issue #1999 - Add telemetry events for the coupon UI

benjamin-cheng:
Related to #1999


app/src/main/java/org/mozilla/rocket/content/CouponAdapter.kt
```diff
@@ -37,10 +37,13 @@ class CouponAdapter(private val listener: ContentPortalListener) : ListAdapter<C
...
         val item = getItem(position) ?: return
         holder.bind(item, View.OnClickListener {
-            TelemetryWrapper.clickOnEcItem(
+            TelemetryWrapper.clickOnPromoItem(
                     pos = position.toString(),
+                    id = item.id,
+                    feed = item.feed,
                     source = item.link.source,
-                    category = item.link.name
+                    category = item.category,
```
cnevinc: now link's name is duplicate with category.
we can't change the schema easily, I'd like to prevent this from happening.
can we re-use the name from category's getter in Coupon class ?

PR #3517, opened 2019-04-24
Issues/3429/fullcheck

mTwTm:
Basic work done: telemetry pending


app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -1233,6 +1323,7 @@ private void showContentPortal() {
         if (contentPanel != null) {
             contentPanel.show(true);
             if (AppConfigWrapper.hasEcommerceCoupons()) {
+                initCouponBanner(getContext());
```
benjamin-cheng: Not a blocker. Just think it would be better to move coupon banner related code into `ContentPortalView`

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -85,6 +91,41 @@ class ContentPortalView : CoordinatorLayout, ContentPortalListener {
         recyclerView?.layoutManager = LinearLayoutManager(context)
         recyclerView?.adapter = couponAdapter
 
+        banner = findViewById(R.id.content_banners)
+        bannerLayoutManager = LinearLayoutManager(context, RecyclerView.HORIZONTAL, false)
+        banner!!.layoutManager = bannerLayoutManager
+        val snapHelper = object : PagerSnapHelper() {
```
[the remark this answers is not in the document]
mTwTm: It was reserved for telemetry but not that we're sure we don't need them yet I'll remove them

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
mTwTm: Had a discussion slightly related to the moving this morning: ContentPortalView is currently treated like a dumb view and has no data loading logic in it so that's why I'm keeping it that way and leaving the loading to activity/fragment. However I should have abstract a class to keep these logics from bloating the HomeFragment. I'll se what I can do later.

PR #3520, opened 2019-04-24
Whitelist strict-mode disk write violation triggered by TabViewProvider.purify()

pharspher:


pharspher:
In my opinion, if we really want to call purify() anytime and anywhere (which indeed is not required according to the current spec), we should find a way to make tab-view provider globally accessible. It doesn't make sense to abstract the creation of tab-view while calling all kinds of implementation specific methods when you want to clear it.

PR #3526, opened 2019-04-24
Support Tab for Content Portal

cnevinc:
This PR
1. Make ContentProtaViewl Child view inflation -> fragment transaction
2. Move Banner Adapter creation to EcFragment
3. Support ViewPager in BottomSheetBehaviour.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -52,6 +52,10 @@ android {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
+
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8
+        }
     }
 
     dexOptions {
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/mozilla/focus/utils/AppConstants.java
```diff
@@ -43,4 +44,8 @@ public static boolean isNightlyBuild() {
         return BuildConfig.FLAVOR_product == FLAVOR_product_NIGHTLY;
     }
 
+    public static boolean isProductionBuild() {
```
cnevinc: we are using focusWebkitReleasefor production build.
and previewWebkitFirebase for nightly build

app/build.gradle
```diff
@@ -52,6 +52,10 @@ android {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
+
+        kotlinOptions {
```
cnevinc: Kotlin need this to support calling static methods on interfaces

PR #3552, opened 2019-04-30
Issue #3549 - Transition life feed newsource to single source

### LINKED ISSUE #3549: Transition life feed newsource to single source
Areas to cover:
1. remote config
2. setting UI (Eng wanted)
https://github.com/mozilla-tw/FirefoxLite/blob/12125ca9ec8f22374fb9db1d089f68fd6d9e0e5a/app/src/main/java/org/mozilla/rocket/widget/NewsSourcePreference.java#L49
This hardcoded options should be only newspoint. 

3. sumo page update  (Bug 1547652 - [SUMO] Removing DainikBhaskar.com)
4. product communication (through push notification, etc)

benjamin-cheng:
Close #3549


app/src/main/java/org/mozilla/rocket/widget/NewsSourcePreference.java
```diff
@@ -9,16 +9,18 @@
...
 import org.mozilla.rocket.content.NewsSourceManager;
 
 public class NewsSourcePreference extends ListPreference {
 
+    // "DainikBhaskar.com" doesn't provide their feed anymore.
     public final static String NEWS_DB = "DainikBhaskar.com";
     public final static String NEWS_NP = "Newspoint";
 
-    @StringDef({NEWS_DB, NEWS_NP})
+    @StringDef({NEWS_NP})
```
mTwTm: This annotation is not used at all, consider using it or removing it. Future addition of this annotation may lead to unwanted crash if user is migrating from before and after the addition of this annotation.

benjamin-cheng:
@tyu good catch for the android-permission-handler version upgrade, I am not intending to modify this file in this PR and I'll remove it from the commit. Thanks!

PR #3553, opened 2019-04-30
Fix crash due to trying to hide the non-exist settings preference

benjamin-cheng:


mTwTm:
checked the preference key and I think the rest usage are all needed fair uses.

PR #3556, opened 2019-05-05
Content portal tabs/fullcheck

cnevinc:
ready for review

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -52,6 +52,10 @@ android {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
+
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8
+        }
     }
 
     dexOptions {
--- a/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
+++ b/app/src/main/java/org/mozilla/focus/home/HomeFragment.java
@@ -163,7 +163,6 @@ public class HomeFragment extends LocaleAwareFragment implements TopSitesContrac
     private static final int SCROLL_PERIOD = 10000;
     private BannerConfigViewModel bannerConfigViewModel;
     final Observer<String[]> homeBannerObserver = this::setUpHomeBannerFromConfig;
-    final Observer<String[]> couponBannerObserver = this::setUpCouponBannerFromConfig;
     private String[] homeBannerconfigArray;
     private String[] couponBannerconfigArray;
     private LottieAnimationView downloadingIndicator;
@@ -333,7 +332,7 @@ public class HomeFragment extends LocaleAwareFragment implements TopSitesContrac
     }
 
     private void initCouponBanner(Context context) {
-        initBanner(context, AppConfigWrapper.getCouponBannerRootConfig(), CURRENT_COUPON_BANNER_CONFIG, bannerConfigViewModel.getCouponConfig(), contentPanel::hideCouponBanner);
+        initBanner(context, AppConfigWrapper.getCouponBannerRootConfig(), CURRENT_COUPON_BANNER_CONFIG, bannerConfigViewModel.getCouponConfig(), null);
     }
 
     private void hideHomeBannerProcedure(Void v) {
@@ -353,7 +352,9 @@ public class HomeFragment extends LocaleAwareFragment implements TopSitesContrac
         // Setup from Network
         if (TextUtils.isEmpty(manifest)) {
             deleteCache(context, cacheName);
-            hideBannerProcedure.accept(null);
+            if (hideBannerProcedure != null) {
+                hideBannerProcedure.accept(null);
+            }
         } else {
             Callback callback = new Callback(context, cacheName, configLiveData);
             new LoadRootConfigTask(callback).execute(manifest, WebViewProvider.getUserAgentString(getActivity()), Integer.toString(SocketTags.BANNER));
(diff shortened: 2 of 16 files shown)


app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -150,6 +178,19 @@ class ContentPortalView : CoordinatorLayout, ContentPortalListener {
         visibility = VISIBLE
         // TODO: if previously is collapsed, collapse here.
         bottomSheetBehavior?.state = BottomSheetBehavior.STATE_EXPANDED
+
+        // content fragment will be added every time when it's displayed. Since back press
+        // will always remove it from fragment manager
+        if (features().size != 1 && !AppConstants.isProductionBuild()) {
+            initContentFragment()
```
mTwTm: Does it make sense that initContentFragment() is delayed but initLegacyContentPortal() is not?

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -308,5 +338,11 @@ class ContentPortalView : CoordinatorLayout, ContentPortalListener {
...
         private const val NEWS_THRESHOLD = 10
         private const val SHOPPINGLINK_GRID_SPAN = 2
+        private const val TAG_CONTENT_FRAGMENT = "TAG_CONTENT_FRAGMENT"
     }
-}
\ No newline at end of file
+}
+
+// helper method to work with FragmentManager
+inline fun Context.inTransaction(func: FragmentTransaction.() -> FragmentTransaction) {
```
mTwTm: Consider place this somewhere else.

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
```diff
@@ -330,7 +328,7 @@ private void initHomeBanner(Context context) {
     }
 
     private void initCouponBanner(Context context) {
-        initBanner(context, AppConfigWrapper.getCouponBannerRootConfig(), CURRENT_COUPON_BANNER_CONFIG, bannerConfigViewModel.getCouponConfig(), contentPanel::hideCouponBanner);
+        initBanner(context, AppConfigWrapper.getCouponBannerRootConfig(), CURRENT_COUPON_BANNER_CONFIG, bannerConfigViewModel.getCouponConfig(), null);
```
mTwTm: consider removing the least parameter, the purpose of it was solely to delegate callback

app/build.gradle
```diff
@@ -52,6 +52,10 @@ android {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
         }
+
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8
```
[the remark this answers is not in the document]
cnevinc: Not every module has Kotlin. So it'll be batter let the library module define it.

app/src/main/java/org/mozilla/focus/home/HomeFragment.java
cnevinc: HomeBanner do need `hideHomeBannerProcedure`, so it's better to keep it.

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -63,40 +66,65 @@ class ContentPortalView : CoordinatorLayout, ContentPortalListener {
...
-            setupViewNews()
+            features.add(TYPE_COUPON)
+        }
+
+        if (AppConfigWrapper.hasEcommerceShoppingLink()) {
+            features.add(TYPE_TICKET)
+        }
+
+        if (AppConfigWrapper.hasNewsPortal()) {
+            // features.add(TYPE_NEWS) // we don't support News in Tab now.
```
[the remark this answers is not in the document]
cnevinc: oh... I think I've made a mistake. I should still allow TYPE_NEWS here. Cause I'll remove it in the `adjust()` method

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
cnevinc: yes. `initContentFragment ()` needs to be called **every time** when we display the ContentPortalView.
While  "initLegacyContentPortal" will inflate the ContentPortalView and its content with HomeFragment. So we don't need to add it again.

cnevinc: fixed

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
cnevinc: fixed

cnevinc:
> ing commit to turn on this feature in production. Before that, we should not merge this PR into master

I'd like to land this and only enable it in Nightly.
While QA verifying in Nightly, I'll add some tests before it goes live.

app/src/main/res/layout/content_tab_coupon.xml
```diff
@@ -31,7 +31,7 @@
             app:layout_constraintBottom_toBottomOf="parent"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toBottomOf="@id/content_banners"/>
+            app:layout_constraintTop_toBottomOf="@id/content_banners "/>
```
benjamin-cheng: Space was inserted in the wrong location. wonder why it can be compiled? However, lint catches it.

PR #3560, opened 2019-05-07
Implement new life feed onboarding

### LINKED ISSUE #3402: Improve conversion from FxLite Home to Life Feed

benjamin-cheng:
Close #3402


app/src/main/java/org/mozilla/rocket/content/LifeFeedOnboarding.kt
```diff
@@ -0,0 +1,64 @@
...
+import android.text.style.ClickableSpan
+import android.view.View
+import org.mozilla.focus.R
+import org.mozilla.focus.activity.InfoActivity
+import org.mozilla.focus.utils.AppConfigWrapper
+import org.mozilla.focus.utils.SupportUtils
+
+object LifeFeedOnboarding {
+
+    private val ONBOARDING_VERSION = 1
```
pharspher: it seems this should be declared as a const val, not important though

PR #3565, opened 2019-05-08
Send revenue-relevant events to adjust

### LINKED ISSUE #3445: Send revenue-relevant events to adjust to optimize revenue
**User story**
As a PMM, I would like to have revenue-relevant events available on adjust to analyze/optimize revenue performance of each acquisition channel.

1. Identify revenue-relevant events
2. Send those events to adjust
3. Verify on adjust in sandbox mode

benjamin-cheng:
Close #3445


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -63,6 +63,8 @@ object TelemetryWrapper {
     private const val SEARCHCLEAR_TELEMETRY_VERSION = "2"
     private const val SEARCHDISMISS_TELEMETRY_VERSION = "2"
 
+    private val AFFILIATE_SITES = listOf("Bukalapak", "Tokopedia")
```
cnevinc: I think "Bukalapak", "Tokopedia" should not be hard-coded.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -1001,6 +1005,10 @@ object TelemetryWrapper {
 
         EventBuilder(Category.ACTION, Method.ADD, Object.TAB, Value.TOPSITE)
                 .queue()
+
+        if (AFFILIATE_SITES.contains(source)) {
```
cnevinc: Please add a comment like "source could be other values" or "we only care about this two source"..

Or maybe just 
```
if (!!TextUtils.isEmpty(source)) {
     AdjustHelper.trackEvent(EVENT_CLICK_AFFILIATE_LINK)
}

```

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
benjamin-cheng: Fixed

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
benjamin-cheng: Fixed

PR #3567, opened 2019-05-08
Give user a way to share us the id for user testing

### LINKED ISSUE #3530: Give user a way to share us the id for user testing
Per discussion with Yvonne, we'd like to give users a way to share us their ids.

I think it's possible via `TelemetryHolder.get().getClientId()` but we'd need to verify it's the correct id. (the formats are the same per what I see.)

benjamin-cheng:
Close #3530


benjamin-cheng:
@cnevinc I already add new lines to separate the fcm token and telemetry id. Is that enough? Or do you prefer other format?

```
[fcm_token]


[telemetry_id]
```

PR #3588, opened 2019-05-13
Content tabs

cnevinc:
pre-work for tabbed-news


benjamin-cheng:
How about moving ktlint related commits from this PR? We may need more discussion about upgrading ktlint.

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
```diff
@@ -50,6 +44,8 @@ class ContentPortalView : CoordinatorLayout, ContentPortalListener {
     private var newsAdapter: NewsAdapter<NewsItem>? = null
     private var newsListLayoutManager: LinearLayoutManager? = null
 
+    private val contentFeatureUseCase = ContentFeaturePresenter()
```
benjamin-cheng: `ContentFeaturePresenter` should rename to `ContentFeatureUseCase` or `ContentFeatureModel` since its role is not like what presenter do.

app/src/main/java/org/mozilla/rocket/content/HomeFragmentViewState.kt
```diff
@@ -16,13 +16,15 @@ object HomeFragmentViewState {
 
     @JvmStatic
     fun reset() {
-        state = TYPE_DEFAULT
+        state =
```
benjamin-cheng: Should keep in the same line

app/src/main/java/org/mozilla/rocket/content/HomeFragmentViewState.kt
```diff
@@ -16,13 +16,15 @@ object HomeFragmentViewState {
...
+            TYPE_DEFAULT
         lastScrollPos = null
     }
 
     @JvmStatic
     fun lastOpenNews() {
-        state = HomeFragmentViewState.TYPE_NEWS
+        state =
```
benjamin-cheng: Should keep in the same line

app/src/main/java/org/mozilla/rocket/content/ContentPortalView.kt
cnevinc: I agree it's not a presenter. It didn't interact with the view and model. And firstly I also name it ContentFeatureUseCase. But at the second thought, it's not like an Interactor/UseCase either. It doesn't interact with the presenter and the controller. It looks much like a model but since I didn't have any architecture here, I don't want to call it model. I'll just use the name `ContentFeature`, and change the name if required in the future.

benjamin-cheng: Thanks for the detail.

PR #3592, opened 2019-05-14
Strings of news category

Jacklin-ux:


Conjuror:
l10n patch is ready. Once the patch is landed on Firefox Lite master, I'll submit the l10n request.

https://github.com/Conjuror/zerda-android-l10n/tree/export-2019-05-14

PR #3596, opened 2019-05-15
Move news UI from ContentPortalView to NewsFragment

benjamin-cheng:


app/src/main/java/org/mozilla/rocket/content/news/NewsPresenter.kt
```diff
@@ -18,8 +18,7 @@ interface NewsViewContract {
     fun updateNews(items: List<NewsItem>?)
 }
 
-class NewsPresenter(private val newsViewContract: NewsViewContract) :
-    ContentPortalView.NewsListListener {
+class NewsPresenter(private val newsViewContract: NewsViewContract) : NewsListListener {
```
cnevinc: nit: please move parameters in new line

benjamin-cheng: Since it can fit on a single line, it may not need to move parameters in another line.
https://developer.android.com/kotlin/style-guide#functions

PR #3601, opened 2019-05-16
fix Kotlin JVM 1.8 compatible issue

proevan:
to workaround the Kotlin JVM 1.8 compatible issue under Android SDK 23 and lower version: https://youtrack.jetbrains.com/issue/KT-31027


pharspher:
To me, visibility map seems more like a SparseArray instead of map. And it also avoids the key hashing operation here, thus there's no need to override hashcode()

PR #3604, opened 2019-05-17
More refactoring on News with Categories

cnevinc:
Previously content portal only have news, then have e-c.
This PR
1.  align the object name for those changes.
2. Move chrome related code to chrome package. Since it's not content related


app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -30,7 +30,7 @@
 /**
  * A wrapper around FirebaseContract. It's job:
  * 1. Interact with FirebaseContract and initialize Firebase
- * 2. Provide helper methods for Firebase related features
+ * 2. Provide helper methods for Firebase related eCommerceFeatures
```
benjamin-cheng: It's not specific to eCommerceFeatures I guess.

app/src/main/java/org/mozilla/rocket/content/portal/ContentFeature.kt
```diff
@@ -3,7 +3,7 @@ package org.mozilla.rocket.content.portal
 import org.mozilla.focus.utils.AppConfigWrapper
 
 /**
- * A presenter to store the logic of providing content portal features
+ * A presenter to store the logic of providing content portal eCommerceFeatures
```
benjamin-cheng: refactoring error

PR #3610, opened 2019-05-20
Fix the crash when no bookmark button on the menu bottom bar

proevan:
- Fix the crash when no bookmark button on the menu bar

Steps:
1. config the bottom bar with a bookmark button, and with no bookmark button on the menu bottom bar.
2. click bookmark button while browsing web page
3. app crash

- added `focusedSessionUrl` as a LiveData in `ChromeViewModel`

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -187,6 +187,9 @@ public class MainActivity extends BaseActivity implements FragmentListener,
         super.onCreate(savedInstanceState);
         chromeViewModel = Inject.obtainChromeViewModel(this);
         downloadIndicatorViewModel = Inject.obtainDownloadIndicatorViewModel(this);
+        BookmarkViewModel.Factory factory = new BookmarkViewModel.Factory(
+                BookmarkRepository.getInstance(BookmarksDatabase.getInstance(this)));
+        bookmarkViewModel = ViewModelProviders.of(this, factory).get(BookmarkViewModel.class);
 
         asyncInitialize();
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -227,15 +230,9 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        BookmarkViewModel.Factory factory = new BookmarkViewModel.Factory(
-                BookmarkRepository.getInstance(BookmarksDatabase.getInstance(this)));
-
-        bookmarkViewModel = ViewModelProviders.of(this, factory).get(BookmarkViewModel.class);
 
         PreferenceManager.getDefaultSharedPreferences(this)
                 .registerOnSharedPreferenceChangeListener(this);
-
-        downloadIndicatorViewModel = Inject.obtainDownloadIndicatorViewModel(this);
```
pharspher: Where did this go?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -1288,6 +1288,7 @@ public void onSecurityChanged(@NonNull Session session, boolean isSecure) {
 
         @Override
         public void onUrlChanged(@NonNull Session session, @Nullable String url) {
+            chromeViewModel.onFocusedSessionChanged(url);
```
pharspher: calling onFocusedSEssionChanged() in onUrlChanged() seems weird to me. Maybe there's a better naming for onFocusedSessionChanged

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
proevan: it's duplicate

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
proevan: updated to onFocusedUrlChanged

PR #3611, opened 2019-05-20
Fix tab counter animation

proevan:
- fix tab counter animation
- observe tab count in ChromeViewModel

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -80,7 +80,6 @@ import org.mozilla.focus.widget.AnimatedProgressBar;
 import org.mozilla.focus.widget.BackKeyHandleable;
 import org.mozilla.focus.widget.FindInPage;
 import org.mozilla.focus.widget.FragmentListener;
-import org.mozilla.focus.widget.TabRestoreMonitor;
 import org.mozilla.permissionhandler.PermissionHandle;
 import org.mozilla.permissionhandler.PermissionHandler;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -558,8 +558,8 @@ private void setupBottomBar(View rootView) {
             bottomBarItemAdapter.setNightMode(isNightMode);
             return Unit.INSTANCE;
         });
-        LiveDataExtensionKt.nonNullObserve(chromeViewModel.getTabCount(), this, changedEvent -> {
-            bottomBarItemAdapter.setTabCount(changedEvent.getCount(), changedEvent.getWithAnimation());
+        LiveDataExtensionKt.nonNullObserve(chromeViewModel.getTabCount(), this, count -> {
+            bottomBarItemAdapter.setTabCount(count, true);
```
benjamin-cheng: I just check the previous implementation, only the one in `onSessionCountChanged` should show tab count change animation.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -1569,7 +1569,7 @@ public void onSessionAdded(@NonNull final Session tab, @Nullable final Bundle ar
         @Override
         public void onSessionCountChanged(int count) {
             if (isTabRestoredComplete()) {
-                chromeViewModel.onTabCountChanged(count, true);
+                chromeViewModel.onTabCountChanged(count);
```
benjamin-cheng: This should have tab count change animation.

Issue #3620, opened 2019-05-22
[top crash] ContentPortalView.kt line 111 


chianger:
Happens since v1.6.0, see:

https://console.firebase.google.com/project/zerda-dcf76/crashlytics/app/android:org.mozilla.rocket/issues/5cdccb32f8b88c296308a719?time=last-ninety-days&sessionId=5CE3ACCC015C0001650C0E4F0D1C6AD6_DNE_0_v2

chianger:
close as #3620 landed.

Daisy-pliu:
This crash not happened on 1.6.2 so verify ✅

PR #3621, opened 2019-05-22
Hotfix for IllegalStateException when init content portal fragments

cnevinc:


chianger:
for #3620

cnevinc:
will use #3622

PR #3627, opened 2019-05-22
Fix background color issue when launching private mode

pharspher:


cnevinc:
This PR does two things.
1. Fix the background color issue when launching private mode
2. Fix the issue when counting leaving private mode, only HomeScreen is counted.
For 1. I suggest:
- adding a darker or grey background to avoid the contract ( white(launcher)->(dark)private)
- fix the problem of launch time and make launcher activity transparent
- if we want to start private mode directly, use an alias to avoid breaking the shortcut if we change the architecture in the future.

For 2. I feel like this will make the pop-up / usage much lower than we expected. 

The code looks good to me so I gave this a PR an r+, and let @pharspher and @Jacklin-ux  decide what to do best.

app/src/main/java/org/mozilla/rocket/privately/ShortcutUtils.kt
```diff
@@ -11,23 +11,21 @@ import android.support.v4.content.pm.ShortcutInfoCompat
...
 import android.support.v4.graphics.drawable.IconCompat
 import org.mozilla.focus.R
-import org.mozilla.focus.utils.AppConstants
 import org.mozilla.rocket.component.LaunchIntentDispatcher
 
 class ShortcutUtils {
     companion object {
         private const val SHORTCUT_ID = "pb_shortcut"
 
         fun createShortcut(context: Context) {
```
cnevinc: could you rename this to `createPrivateShortcut ` >

app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -269,9 +269,10 @@ class PrivateModeActivity : BaseActivity(),
     }
 
     private fun checkShortcutPromotion(continuation: () -> Unit) {
+        val isHomeScreen = screenNavigator.topFragment?.tag == ScreenNavigator.HOME_FRAGMENT_TAG
```
cnevinc: add a comment to explain why it's needed. since the boolean parameter is always confusing.

app/src/main/java/org/mozilla/rocket/privately/ShortcutUtils.kt
```diff
@@ -37,8 +38,23 @@ class ShortcutUtils {
                     .setIcon(icon)
                     .build()
 
+            // Known issue: when there is already an existing shortcut, intendSender will be called
+            // immediately when system shortcut dialog is shown. i.e. Home will be brought up, and then
+            // show the system shortcut dialog. UX currently agreed with this behavior since it's
+            // not often the user will add shortcut again.
+            val homeIntent = Intent(Intent.ACTION_MAIN)
```
cnevinc: nit: extra this block and reuse it.

pharspher:
Separate the original commits into another two PRs #3629, #3630

app/src/main/java/org/mozilla/rocket/privately/ShortcutUtils.kt
pharspher: > 1. Fix the background color issue when launching private mode
> 
> * adding a darker or grey background to avoid the contract ( white(launcher)->(dark)private)
> * fix the problem of launch time and make launcher activity transparent
> * if we want to start private mode directly, use an alias to avoid breaking the shortcut if we change the architecture in the future.

After the discussion with @Jacklin-ux, it might be the unsmooth animation that makes the transition weird, not the background color. For now I'll directly launch the alias activity of PrivateModeActivity. After improving the transition animation, I'll change the alias target back to the RocketLauncherActivity.

PR #3633, opened 2019-05-23
Dynamic private bottom bar

proevan:
Use `BottomBar` in private mode and make it configurable from the remote config as well.
* This PR only make it support the default items, and will do the rest items implementation in the next PR.


app/src/main/java/org/mozilla/rocket/privately/browse/BrowserFragment.kt
```diff
@@ -81,18 +86,19 @@ class BrowserFragment : LocaleAwareFragment(),
...
     private lateinit var trackerPopup: TrackerPopup
 
-    private var isLoading: Boolean = false
-
     private var systemVisibility = ViewUtils.SYSTEM_UI_VISIBILITY_NONE
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        chromeViewModel = Inject.obtainChromeViewModel(activity).apply {
+            onRestoreTabCountCompleted()
```
pharspher: why do we need to call onRestoreTabCountCompleted() here in private mode?

proevan: since I think it's not using in private mode, so I want make the `isTabRestoredComplete` to be `ture` from the start when in the private mode

proevan: removed the code that calling `onRestoreTabCountCompleted()` in private mode

PR #3642, opened 2019-05-27
Support more actions for the bottom bar in private mode

proevan:
Previous PR #3633 has implemented the default bottom bar buttons.
Support more actions in this PR.

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
+++ b/app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
@@ -12,6 +12,7 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
+import android.net.Uri
 import android.os.Bundle
 import android.support.annotation.CheckResult
 import android.support.v4.app.Fragment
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -110,14 +119,59 @@ class PrivateModeActivity : BaseActivity(),
 
     override fun applyLocale() {}
 
+    private fun observeChromeAction() {
+        chromeViewModel.showUrlInput.observe(this, Observer { url ->
+            if (!supportFragmentManager.isStateSaved) {
+                screenNavigator.addUrlScreen(url)
+            }
+        })
+        chromeViewModel.pinShortcut.observe(this, Observer {
```
pharspher: do we need this?

proevan: comment added

app/src/main/java/org/mozilla/rocket/privately/PrivateModeActivity.kt
```diff
@@ -111,14 +120,60 @@ class PrivateModeActivity : BaseActivity(),
 
     override fun applyLocale() {}
 
+    private fun observeChromeAction() {
+        // Reserve to handle more chrome actions for the bottom bar A/B testing
```
proevan: here

Issue #3137, opened 2019-01-31
Migrate from support libraries to AndroidX


pocmo:
Migration tracking meta issue: https://github.com/mozilla-mobile/android-components/issues/1903

For details please read the linked meta issue.

TL;DR: We want to switch from the support libraries to the AndroidX libraries since that is where new features are getting developed. We need downstream app projects to migrate first before we can migrate up the dependency tree (Android Components, Application Services, GeckoView, ..).

Daisy-pliu:
This refactoring has huge impact so a full run test is requested after PR landing.

PR #3646, opened 2019-05-28
Upgrade Android Components

Holidaybear:
- Upgrade Android Components to 0.51.0
- Add `lib-fetch-httpurlconnection` because `HttpURLConnectionTelemetryClient` was removed
- Modify UrlInputFragment because `DomainAutoCompleteProvider` was deprecated & OnFilterListener 2nd param was removed
- Modify UrlInputFragment layout in order to pass UI test (SearchFieldTest)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.kt
@@ -18,7 +18,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.webkit.URLUtil
 import android.widget.TextView
-import mozilla.components.browser.domains.DomainAutoCompleteProvider
+import mozilla.components.browser.domains.autocomplete.ShippedDomainsProvider
 import mozilla.components.ui.autocomplete.InlineAutocompleteEditText
 import org.mozilla.focus.Inject
 import org.mozilla.focus.R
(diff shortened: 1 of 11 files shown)


Holidaybear:
related to #3137

cnevinc:
lgtm

app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.kt
```diff
@@ -188,11 +188,14 @@ class UrlInputFragment : Fragment(), UrlInputContract.View, View.OnClickListener
...
-            when {
-                it.isEmpty() || !URLUtil.isValidUrl(urlView.text.toString()) -> urlView.text.toString()
-                else -> it
+        val input = urlView.autocompleteResult?.let { result ->
+            if (result.text.isEmpty() || !URLUtil.isValidUrl(urlView.text.toString())) {
+                urlView.text.toString()
+            } else {
+                result.text
             }
+        } ?: run {
```
pharspher: run() seems unnecessary here?

components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/TabViewEngineObserver.kt
```diff
@@ -83,7 +83,7 @@ class TabViewEngineObserver(
 
     override fun onExternalResource(
         url: String,
-        fileName: String?,
+        fileName: String,
```
pharspher: it seems the caller of this method get all parameters from a Download object, which declares its name field as nullable, is it safe here to change it to non-optional type?

app/src/main/java/org/mozilla/focus/urlinput/UrlInputFragment.kt
Holidaybear: If `urlView.autocompleteResult == null`, will run this statement.

components/feature/tabs/src/main/java/org/mozilla/rocket/tabs/TabViewEngineObserver.kt
Holidaybear: After upgrading android-components, they changed its `fileName` field to non-null from `EngineSession.Observer`

PR #3651, opened 2019-05-28
Apply hotfix from 1.6.2 to master

cnevinc:
Now 1.6.2 has shipped. We should apply the hotfix from 1.6.2


cnevinc:
already had a r+ in #3622.
will carry over that r+ and land it on master once it passes the CI

PR #3654, opened 2019-05-29
Add preference to record that private shortcut has been created

pharspher:


chianger:
adding note here so that we'll remember what this impact is:
if user add PB shortcut to homescreen thru the setting UI, he/she will no longer get the prompt to do so.

Issue #3665, opened 2019-05-31
News can not display


Daisy-pliu:
**Describe the bug**
User in India can not see news as it is  loading forever

**Steps To Reproduce**
1. vpn in India
2. fresh install  NIghtly 1.6.2 (12538)
3. skip onboarding 
4. swipe [up] to see [news]

**Expected behavior**
see [lastest news]

**Actual behavior**
whole page is loading forever

**Screenshots/videos**
![Screenshot_20190531-173040](https://user-images.githubusercontent.com/44990878/58696528-59848400-83ca-11e9-8835-6e1ad9171cab.png)

**Device info (please complete the following information):**
 - Device: pixel 2 (9.0) webview beta:  75.0.3770.40
 - Device: oppo (7.1.1) webview stable: 73.0.3683.90

**Reproduce rate**
6/6

**Additional context**
This issue not happen on release version 1.6.2 (12322)

benjamin-cheng:
I've updated the remote config on Firebase nightly that should resolve this issue.

chianger:
Closed and to be verified.

Daisy-pliu:
verify: ✅
version : 1.7.0(12623) RC
test device: samsung 8.0 webview : 74.0.3729.157 and webview beta: 75.0.3770.67
test case: TC0198~TC0202
Note: Test on 1.6.2 (12322) RC as well good to go.

PR #3667, opened 2019-06-01
Apply the same life feed provider remote config accross different app version

benjamin-cheng:
I've created a new remote config (str_news_providers) for the news with category API endpoint setting. However, I think it would increase the complexity to manage news API endpoints. So I try to merge the new config back to the existing one(life_feed_providers) and not to impact the legacy user who does not update to the new news enhancement. 

Related to #3665
TODO:
Delete `str_news_providers` from remote config of dev and nighly

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
+++ b/app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
@@ -15,15 +15,18 @@ import org.mozilla.rocket.content.ecommerce.data.Coupon;
 import org.mozilla.rocket.content.ecommerce.data.CouponKey;
 import org.mozilla.rocket.content.ecommerce.data.ShoppingLink;
 import org.mozilla.rocket.content.ecommerce.data.ShoppingLinkKey;
+import org.mozilla.rocket.content.news.data.NewsProviderConfig;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class AppConfigWrapper {
     static final int SURVEY_NOTIFICATION_POST_THRESHOLD = 3;
     static final boolean PRIVATE_MODE_ENABLED_DEFAULT = true;
     static final boolean LIFE_FEED_ENABLED_DEFAULT = false;
-    static final String STR_NEWS_PROVIDERS_DEFAULT = "";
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
```diff
@@ -33,6 +35,11 @@
     private static final boolean SURVEY_NOTIFICATION_ENABLED = false;
     static final int DRIVE_DEFAULT_BROWSER_FROM_MENU_SETTING_THRESHOLD = 2;
 
+    private static final Map<String, String> lifeFeedProviderConfigNameMapping = new HashMap<>();
+    static {
```
cnevinc: please add some comment to it. I also feel it's like you are doing a data migration, you'll probably need a uni test for this to avoid future breakage.

app/src/main/java/org/mozilla/focus/utils/FirebaseHelper.java
```diff
@@ -60,7 +60,7 @@
     static final String FIRST_LAUNCH_TIMER_MINUTES = "first_launch_timer_minutes";
     static final String FIRST_LAUNCH_NOTIFICATION_MESSAGE = "first_launch_notification_message";
     static final String ENABLE_LIFE_FEED = "enable_life_feed";
-    static final String STR_NEWS_PROVIDERS = "str_news_providers";
+    static final String STR_NEWS_PROVIDERS = "life_feed_providers";
```
cnevinc: how about making the var the same name liks LIFE_FEED_PROVIDERS?

app/src/main/java/org/mozilla/focus/utils/AppConfigWrapper.java
cnevinc: Or at least create an issue and make the unit test the priority for next sprint.

cnevinc: I had an impression we previously add the priority of how we can override news source. Maybe it's worth looking when you want to add a default value.

benjamin-cheng: I've filed the issue to add unit test #3670

PR #3669, opened 2019-06-03
Create menu dialog

proevan:
Create a class `MenuDialog` and move the code that relate to menu into this class

This PR is based on the PR #3652
please review starting from the commit: `Created MenuDialog that set up menu bottom bar and display menu items`

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -30,11 +30,8 @@ import android.support.v4.app.Fragment;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.NotificationManagerCompat;
 import android.support.v4.content.LocalBroadcastManager;
-import android.support.v4.content.pm.ShortcutManagerCompat;
-import android.support.v7.app.AppCompatDialog;
 import android.text.TextUtils;
 import android.view.View;
-import android.view.View.OnLongClickListener;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/rocket/extension/Extension.kt
```diff
@@ -0,0 +1,22 @@
+package org.mozilla.rocket.extension
```
pharspher: Would it be better to rename to something like ContextExtenstion or ContextExt, etc? Or this file is planned to contain extensions of all kinds of classes?

app/src/main/java/org/mozilla/rocket/extension/Extension.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
+import android.support.v4.app.FragmentActivity
+
+
+fun Context.toActivity(): Activity = when {
+    this is Activity -> this
+    this is ContextWrapper -> this.baseContext.toActivity()
```
pharspher: what about nested context wrapper whose ancestor is an Activity?

app/src/main/java/org/mozilla/rocket/extension/Extension.kt
proevan: updated to ContextExtension

proevan:
merge into PR #3682

PR #3671, opened 2019-06-03
Change Screengrab run on debug build

Daisy-pliu:
1. Screengrab needs debug build instead of firebase build after we change code base on UI testing.
2. Screengrab need to add permission  "android.permission.CHANGE_CONFIGURATION"  (refer: https://docs.fastlane.tools/actions/screengrab/) on Debug build.
3. app_package_name="org.mozilla.rocket.debug.pliu" is hard coded as debug build apply Suffix like ".debug." + userName. Any suggestion to this is welcomed. (I checked the document seems no customized parameter can send through command line.)

### DIFF
--- a/fastlane/Screengrabfile
+++ b/fastlane/Screengrabfile
@@ -1,12 +1,13 @@
 
-#locales ['en','hi','in','jv','su','th','tl','zh-TW','vi']
+#locales ['en','hi','in','jv','su','th','tl','zh-TW','vi','ta','kn','ml']
 locales ['vi','zh-TW']
 clear_previous_screenshots true
 exit_on_test_failure false
 test_instrumentation_runner  'org.mozilla.focus.test.runner.ScreenshotTestRunner'
-app_package_name 'org.mozilla.rocket.debug.firebase'
+# hardcode on debug build application_id as fastlane no customized input parameter
+app_package_name 'org.mozilla.rocket.debug.pliu'
 use_tests_in_packages ['org.mozilla.focus.screengrab']
 #use_tests_in_classes ['org.mozilla.focus.screengrab.MyShotScreenshot']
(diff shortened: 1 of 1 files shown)


fastlane/Screengrabfile
```diff
@@ -1,12 +1,12 @@
 
-#locales ['en','hi','in','jv','su','th','tl','zh-TW','vi']
+#locales ['en','hi','in','jv','su','th','tl','zh-TW','vi','ta','kn','ml']
 locales ['vi','zh-TW']
 clear_previous_screenshots true
 exit_on_test_failure false
 test_instrumentation_runner  'org.mozilla.focus.test.runner.ScreenshotTestRunner'
-app_package_name 'org.mozilla.rocket.debug.firebase'
+app_package_name 'org.mozilla.rocket.debug.pliu'
```
cnevinc: maybe add a comment or update the wiki so the future maintainer could know why it's needed.

app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,14 @@
...
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    package="org.mozilla.focus">
+
+    <!--For fastlane usage-->
+    <!-- Allows changing locales -->
+    <uses-permission
+        android:name="android.permission.CHANGE_CONFIGURATION"
```
cnevinc: I'm a little concern about in the future, the developer will have this permission and break some features without notice. I can think of two work around
1. have another build type for screengrab
2. manually paste or programmatically generate this file when we want to run screengrab jobs.

Previously we had a bug that we want to run UI tests on private mode activity so we change its process to be the same as the testing app. And the developers didn't aware of that in dev build so had a strange regression.

PR #3672, opened 2019-06-03
Upgrade gradle plugin & kotlin

Holidaybear:
- Upgrade gradle plugin to 3.4.2
- Upgrade kotlin to 1.3.41
- Modify invalid API caused by upgrading gradle
- Modify warning UI layout


cnevinc:
This is the work I have done to achieve the same goals.
 But I'm stuck on other tasks now. You can help rebase and merge it if you like

https://github.com/mozilla-tw/FirefoxLite/pull/3503

Holidaybear:
looks like we need to keep it pending until AndroidX migration is done
root cause: Robolectric 4.0 above need to switch `androidx.test`
ref: http://robolectric.org/migrating/

Issue #3676, opened 2019-06-05
Configurable menu grid items


proevan:
Since I'm doing the code base clean-up. I want to isolate the Menu feature from the MainActivity in following steps:
1. Isolate the code that relates to Menu out from MainActivity
2. Make the Menu items configurable

proevan:
PR #3652

chianger:
#3682

Daisy-pliu:
Verify: ❌
Test steps:
1. remote config on nightly as [{"type":0},{"type":2},{"type":1},{"type":3},{"type":4},{"type":5},{"type":6},{"type":7},{"type":8},{"type":9},{"type":10},{"type":11}]
2. fresh install nightly
3. launch app
4. skip onboarding 
4. tap [menu]

Test version: 1.8.0 (13041)
Test result: crash
Test device: Oppo 7.1.1
Note: crash happens after  fresh install. if user continue use app and tap [menu] it is fine.

proevan:
> Verify: ❌
> Test steps:
> 
>     1. remote config on nightly as [{"type":0},{"type":2},{"type":1},{"type":3},{"type":4},{"type":5},{"type":6},{"type":7},{"type":8},{"type":9},{"type":10},{"type":11}]
> 
>     2. fresh install nightly
> 
>     3. launch app
> 
>     4. skip onboarding
> 
>     5. tap [menu]
> 
> 
> Test version: 1.8.0 (13041)
> Test result: crash
> Test device: Oppo 7.1.1
> Crash report : [crash LiveDataExtension.kt line 18 ](https://console.firebase.google.com/u/0/project/rocketnightly/crashlytics/app/android:gro.allizom.zelda.beta/issues/f61909231b35b4769258de83da933440?time=last-seven-days&sessionId=5D19842C036D0001555AD438E2856CEB_DNE_0_v2)
> 
> Note: crash happens after fresh install. if user continue use app and tap [menu] it is fine and the grid setting are good as exepected.

fixed in #3745

Daisy-pliu:
refer :https://github.com/mozilla-tw/FirefoxLite/wiki/Remote-Config#str_menu_items
verify: ✅
Test step:[step](https://github.com/mozilla-tw/FirefoxLite/issues/3676#issuecomment-507107809)
Test version: 1.8.0 (13607)
Test result:  see correct menu grid
Test device: Oppo 7.1.1

Daisy-pliu:
[...]

PR #3682, opened 2019-06-05
Make Menu grid items configurable by remote configs

proevan:
#3676

document for remote configs: https://github.com/mozilla-tw/FirefoxLite/wiki/Remote-Config#str_bottom_bar_items


app/src/main/java/org/mozilla/rocket/content/view/MenuLayout.kt
```diff
@@ -0,0 +1,110 @@
...
+        abstract fun createView(context: Context): View
+
+        open class TextImageItem(
+            type: Int,
+            private val textResId: Int,
+            private val drawableResId: Int,
+            private val tintResId: Int
+        ) : MenuItem(type) {
+            override fun createView(context: Context): View {
+                return LinearLayout(context).apply {
```
pharspher: Is there any reason to create the layout programmatically instead of using XML?

proevan: updated to use xml to inflate views

PR #3695, opened 2019-06-14
Fix the News setting telemetry

cnevinc:


app/src/main/java/org/mozilla/rocket/content/news/NewsSettingFragment.kt
```diff
@@ -41,6 +41,13 @@ class NewsSettingFragment : PreferenceFragmentCompat() {
         }
     }
 
+    override fun onPause() {
+        super.onPause()
+        categoryPreference?.NewsCatSettingCatAdapter()
```
benjamin-cheng: Why have to new `NewsCatSettingCatAdapter` instance here?

PR #3701, opened 2019-06-17
Upgrade kotlin to  1.3.31 and coroutine to 1.3.0-M1

pharspher:


benjamin-cheng:
Kotlin 1.3.40 was relased
https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/

PR #3765, opened 2019-07-05
Fix Turbo mode doesn't sync in Settings page and menu

### LINKED ISSUE #3762: Turbo mode doesn't sync in Settings page and menu
**Describe the bug**
When a user turns off Turbo mode, the Turbo mode icon in menu is still on. Vice versa.

**Steps To Reproduce**
1. Launch Firefox Lite
2. Tap on Menu
3. Tap on Settings
4. Disable Turbo mode
5. Tap back and trigger menu again

**Expected behavior**
5. Turbo mode is disabled

**Actual behavior**
5. Turbo mode is disabled but the icon is still in blue (on)

**Screenshots/videos**
N/A

 - Device: Redmi 5 plus
 - OS: Android 8.1.0
 - Webview 75.0.3770.101
 - Version 1.8.0 (13143)

**Reproduce rate**
5/5

**Additional context**
This is regresison which is not there in 1.8.0 (13128)

proevan:
to fix #3762

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/preference/SharedPreferenceLiveData.kt
+++ b/app/src/main/java/org/mozilla/rocket/preference/SharedPreferenceLiveData.kt
@@ -1,0 +1,74 @@
+package org.mozilla.rocket.preference
+
+import android.arch.lifecycle.LiveData
+import android.content.SharedPreferences
+
+abstract class SharedPreferenceLiveData<T>(
+    val sharedPrefs: SharedPreferences,
+    val key: String,
+    private val defValue: T
+) : LiveData<T>() {
+
+    private val preferenceChangeListener = SharedPreferences.OnSharedPreferenceChangeListener { _, key ->
+        if (key == this.key) {
+            value = getValueFromPreferences(key, defValue)
+        }
+    }
+
+    abstract fun getValueFromPreferences(key: String, defValue: T): T
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/rocket/preference/SharedPreferenceLiveData.kt
```diff
@@ -0,0 +1,74 @@
...
+}
+
+class SharedPreferenceIntLiveData(sharedPrefs: SharedPreferences, key: String, defValue: Int) :
+        SharedPreferenceLiveData<Int>(sharedPrefs, key, defValue) {
+    override fun getValueFromPreferences(key: String, defValue: Int): Int = sharedPrefs.getInt(key, defValue)
+}
+
+class SharedPreferenceStringLiveData(sharedPrefs: SharedPreferences, key: String, defValue: String) :
+        SharedPreferenceLiveData<String>(sharedPrefs, key, defValue) {
+    override fun getValueFromPreferences(key: String, defValue: String): String = requireNotNull(sharedPrefs.getString(key, defValue))
```
pharspher: why do we need to requireNotNull here? If the key doesn't exist, the defValue, which is a non-null type variable, will be returned instead. It seems the requreNotNull here is unnecessary. This is not a merge blocker, just want to know if there's any special reason

PR #3780, opened 2019-07-12
Implement a experiment feature - option to block JavaScript in web pages

benjamin-cheng:


app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -23,6 +22,8 @@
 import android.webkit.WebView;
 import android.webkit.WebViewDatabase;
 
+import androidx.annotation.NonNull;
```
cnevinc: do we need this?

benjamin-cheng: Just a re-sort to follow the alphabetical order.

PR #3794, opened 2019-07-31
Games groundwork

proevan:
device-2019-07-29-180331

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/content/games/BrowserGamesFragment.kt
+++ b/app/src/main/java/org/mozilla/rocket/content/games/BrowserGamesFragment.kt
@@ -9,6 +9,7 @@ import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProviders
 import androidx.recyclerview.widget.LinearLayoutManager
 import kotlinx.android.synthetic.main.fragment_games.list
+import kotlinx.android.synthetic.main.fragment_games.spinner
 import org.mozilla.focus.R
 import org.mozilla.rocket.adapter.AdapterDelegatesManager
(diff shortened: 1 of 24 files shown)


app/src/main/java/org/mozilla/rocket/content/games/BrowserGamesFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        initRecyclerView()
+    }
+
+    private fun initRecyclerView() {
+        adapter = DelegateAdapter(
+            AdapterDelegatesManager()
+        )
+        list.apply {
```
benjamin-cheng: The name `list` is somehow confused me, it's more like a data list. Maybe, we could rename it to `recyclerview` to be more readable.

app/src/main/java/org/mozilla/rocket/content/games/GamesViewModel.kt
```diff
@@ -0,0 +1,17 @@
+package org.mozilla.rocket.content.games
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import org.mozilla.rocket.adapter.DelegateAdapter
+
+class GamesViewModel : ViewModel() {
+
+    val browserGamesState = MutableLiveData<State>().apply { value = State.Idle }
```
benjamin-cheng: Just a thought, we could combine these two livedata into one, like: MutableLiveData<State<List<DelegateAdapter.UIModel>>>()

PR #4367, opened 2019-11-21
[Travel] travel city video applying api

kinse:
Part of #3872


app/src/main/java/org/mozilla/rocket/content/travel/data/YoutubeApiEntity.kt
```diff
@@ -0,0 +1,65 @@
...
+                jsonObject.optString(KEY_LINK),
+                jsonObject.optString(KEY_VIEW_COUNT).toInt(),
+                jsonObject.optString(KEY_COMPONENT_ID),
+                jsonObject.optString(KEY_SOURCE)
+            )
+    }
+}
+
+private fun String.toIntDuration(): Int {
+    val regex = """^PT(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?$""".toRegex()
```
benjamin-cheng: How about using `SimpleDateFormat`?
https://stackoverflow.com/questions/10985848/converting-string-ex-pt20h-to-time-in-android

app/src/main/java/org/mozilla/rocket/content/travel/data/YoutubeApiEntity.kt
```diff
@@ -0,0 +1,65 @@
+package org.mozilla.rocket.content.travel.data
+
+import org.json.JSONObject
+import org.mozilla.rocket.util.toJsonArray
+
+data class YoutubeApiEntity(val videos: List<YoutubeApiItem>) {
```
benjamin-cheng: The API result from back-end might have other video sources, so suggest to have a more general name like `VideoApiEntity`

PR #4410, opened 2019-12-03
Update l10n to Dec. 3rd

### LINKED ISSUE #4390: l10n issue
Following string resource should have 1 parameter in `ru/string.xml`

`<string name="msrp_challenge_details_body_2">Чтобы начать введите свой адрес электронной почты. Убедитесь, что вы вошли в аккаунт, который хотите использовать, до получения вознаграждения.</string>`
`<string name="msrp_challenge_tou">Нажимая «Я участвую», вы соглашаетесь с нашими Условиями использования.</string>`

Conjuror:
fixes #4390


app/src/main/res/values-zh-rTW/strings.xml
```diff
@@ -357,8 +357,6 @@
   <string name="content_home_context_menu_share_link">分享鏈結</string>
   <string name="content_home_error_message">喔喔，有些東西不對勁！</string>
   <string name="private_home_title">進階隱私瀏覽模式</string>
-  <string name="private_home_description_1">現在起，您上網可以保持私密，不會留下上網紀錄、登入密碼，其他使用這台裝置的人也不會知道您曾經開過哪些網站。\n\n我們的進階隱私瀏覽功能，甚至可以%1$s，這是其他大部分隱私模式做不到的。</string>
-  <string name="private_home_description_2">保護您不被廣告商或網站追蹤</string>
```
proevan: why do we remove these 2 strings?

Conjuror: As discussed offline, these are legacy issues we have to deal with. I am suggesting to remove the translation result which is not acquired from Pontoon.

Issue #4421, opened 2019-12-04
New key name for travel vertical (remote config)


chianger:
So that we can pref on thru remote config and make sure only users having the new version see travel vertical.

chianger:
Or we can simply unlock the travel disable_lock for production build.

weslyhuang:
per checked it will be done by @benjamin-cheng via remote config.

Issue #4458, opened 2019-12-09
Show travel vertical for first time onboarding & add its description


Jacklin-ux:
For 1.x and new users, after updating or installing Firefox Lite, they will see onboarding with Travel vertical & its introduction.

proevan:
@Jacklin-ux, please ping me once the introduction string resources is ready, thanks!

Conjuror:
17646

PR #4476, opened 2019-12-11
add travel as a default content hub item and update onboarding descri…

proevan:
…ption

to implement #4458


benjamin-cheng:
Close #4421

PR #4477, opened 2019-12-11
allows to show/copy firebase id for in-app-message debugging usage

proevan:
allows to see the firebaseId in debug page that can help QA to debug In-App-Messaging

device-2019-12-11-174012


proevan:
@Daisy-pliu @Conjuror FYI

PR #4481, opened 2019-12-12
[Travel] Enable travel content hub entry in the production build

### LINKED ISSUE #4421: New key name for travel vertical (remote config)

benjamin-cheng:
Close #4421


benjamin-cheng:
Dup with #4476, close it.

PR #4490, opened 2019-12-13
[Travel] Refine wiki api

### LINKED ISSUE #4469: [Travel] Support multiple languages in Wikipedia cards
**User story**
The current implementation only supports English. As a user, I'd like to see the Wikipedia content which is aligned with my UI language.

**Actions**
- [ ] Tina to ask Wikimedia about the API details
- [ ] Benjamin to test if the API can support the requirements
- [ ] Implement the new API

hartsu:
Close #4469

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/content/travel/data/TravelRepository.kt
+++ b/app/src/main/java/org/mozilla/rocket/content/travel/data/TravelRepository.kt
@@ -1,7 +1,9 @@
 package org.mozilla.rocket.content.travel.data
 
+import android.net.Uri
 import org.mozilla.rocket.content.Result
 import org.mozilla.rocket.content.common.data.ApiEntity
+import java.util.Locale
 
 class TravelRepository(
     private val remoteDataSource: TravelRemoteDataSource,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/rocket/content/travel/ui/TravelCityViewModel.kt
```diff
@@ -115,7 +117,13 @@ class TravelCityViewModel(
                     )
                 }
 
-                val wikiResult = getWiki(Uri.encode(englishName))
+                var wikiName = name
+                val wikiNameResult = getCityWikiName(Uri.encode(name))
```
benjamin-cheng: We should encapsulate this logic into use case layer, so view model doesn't have to know that it should get the normalized name first then query the extract and image.

app/src/main/java/org/mozilla/rocket/content/travel/data/TravelRepository.kt
```diff
@@ -29,14 +31,17 @@ class TravelRepository(
     }
 
     suspend fun getCityWiki(name: String): Result<Wiki> {
-        val resultExtract = remoteDataSource.getCityWikiExtract(name)
-        val resultImage = remoteDataSource.getCityWikiImage(name)
+        val resultName = remoteDataSource.getCityWikiName(name)
```
benjamin-cheng: I think we should also keep repository as simple as possible. It should be use case's responsibility(domain knowledge) to coordinate the call sequences to get the composite result. However, it's not a blocker now, please help to refactor it later. Thanks!

Issue #4535, opened 2019-12-22
Microcopy error


iprithvitharun:
**Describe the bug**
Microcopy message which appears after closing the Private Tab uses present perfect passive. Here, present tense can be used for clear and concise message.

**Steps To Reproduce**

1. Open a private tab.
2. Go to any URL.
3. Hit delete icon at the bottom.
4. See the message. 

**Expected behavior**
Your private browser history is deleted.

**Actual behavior**
Your private browser history has been deleted.

**Screenshots/videos**
Screenshot link - https://imgur.com/a/otOOkGt

**Device info (please complete the following information):**
 - Device: [Redmi Note 7]
 - OS: [e.g. Android 9]

**Reproduce rate**
Every time

Conjuror:
Thank you for the valuable inputs.

weslyhuang:
per discussion, works for us.

iprithvitharun:
Hi,

Can you review the request and approve it?

Prithvi
LinkedIn <https://www.linkedin.com/in/prithvi-tharun/> | Twitter
<https://twitter.com/iprithvitharun>

joechengla:
see https://github.com/mozilla-tw/FirefoxLite/pull/4536

PR #4536, opened 2019-12-22
Microcopy typo is fixed

iprithvitharun:
Issue #4535 

Message changed from present perfect continuous tense to present tense for easy understanding.


iprithvitharun:
@benjamin-cheng Can you review this request? Or can you tag the right person?

Delphine:
I talked with the localization team about this, and we agree that the original string sounds more natural. 
Jeff Beatty (localization-drivers manager) mentioned on Slack: "it’s really a question of what is the voice Taipei uses for the product consistently: passive or active. If active, then it should be “was erased” if passive, “has been erased”".
I don't know who the English copywriter for Firefox Lite is, but Michelle Heubusch might  have more insight about this, if needed.

joechengla:
@iprithvitharun I'd like to thank you for your contribution and suggestion here. However, per internal discussion, the team will stick to the current string and tone. Please do not get discouraged by this. Your input is very valuable to us. Thank you

PR #4605, opened 2020-01-06
String update 1/6

youwenliang:
Change Game vertical's add shortcut from "press" to "long press"


benjamin-cheng:
@Conjuror please help to review and merge the string, thanks!

PR #4606, opened 2020-01-07
Finish activity if private mode browsing is from external

mxlius:


proevan:
seems like it still needs to go back twice to leave private mode. This PR need to be adjusted

PR #4612, opened 2020-01-08
String update 1/8

youwenliang:
Personalized News strings


benjamin-cheng:
@youwenliang As we discussed, please add more strings:
`Choose Your Preferred \nNews Language`
`Personalize news experience`

PR #4737, opened 2020-02-21
Move customized resources to standalone module

### LINKED ISSUE #4816: Use Module for customized resources
**Describe the bug**
With customized resources, we can have a better workflow for localizing strings without pending on Pontoon.

mxlius:
Would you help to update the customize string script ```./tools/l10n/attach-local-strings.py```, please?

fixed #4816


Conjuror:
↑ Force push to update l10n strings

Issue #4617, opened 2020-01-09
As a user, if no result for Travel Discovery search, no response on search button


Conjuror:
**Describe the bug**
As title, if Travel discovery is set as default, users cannot user Google to search cities.

**Reproduce steps**
1. Launch Firefox Lite
2. Tap on [Travel]
3. Tap on [search bar] and search for a city
4. Tap on a city to detail view
5. Tap on [Back key] to exit detail view
6. Tap on [Set as default]
7. Search a city name which has no response from search API, i.e. "大鵬灣"

Expected Result:
7. Search by Google

Actual result:
7. No response

Conjuror:
Related to #4484

Conjuror:
Concluded, we need a page for no search result.

Here's another reproduce steps
**Reproduce Steps**
1. Launch Firefox Lite
2. Tap on [Travel]
3. Tap on [Search bar]
4. Type "大鵬灣"
5. Press [Search/Enter]

Expected Result:
5. Search with Google

Actual Result:
5. No response and no information

tinahsieh:
UX spec: https://mozilla.invisionapp.com/share/EWSX97EB69J#/404999237_6-1_No_Search_Results

The copy string is still under review. I'll be updated on both spec and .xml when it's ready.

TLHuang:
The visual spec is ready, please find it here https://zpl.io/bejMG0p

Conjuror:
18889

PR #4754, opened 2020-02-25
[Travel] copy strings for no search result state

tinahsieh:
1 string change: 
- Explore a city -> EXPLORE A CITY

2 strings added:
- Hmm. No results found.
- SEARCH ON %1$s

Related to bug #4617

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -940,8 +940,6 @@
     <string name="travel_empty_state_text_2" translatable="false">- Susan Sontag</string>
     <!--button for entering the search page-->
     <string name="travel_empty_state_button">Explore a city</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -940,6 +940,8 @@
     <string name="travel_empty_state_text_2" translatable="false">- Susan Sontag</string>
     <!--button for entering the search page-->
     <string name="travel_empty_state_button">Explore a city</string>
+    <!--button for entering the search page-->
+    <string name="travel_empty_state_button_new">EXPLORE A CITY</string>
```
benjamin-cheng: Don't have to add this new string. We could make the string all capital in code.

PR #4764, opened 2020-02-26
Server Push Android Work - support data message

cnevinc:
I first thought making push payload as a POJO for server and client could be a good idea.
But FCM SDK already parses the data payload into HashMap.
And we need Bundle as input data for WorkManager to schedule when to display the notification.
So I'll just reuse current architecture.

For extracting the methods.
Normally it's a bad idea to mix the Context of a Service and the Application Context.
But here we only use it for 
1. getting WorkManager
2. getting String Resources, 
3. Building the Notification

So it should be fine


cnevinc:
related:  https://github.com/mozilla-tw/FirefoxLite/issues/4659

proevan:
the updates LGTM

Issue #4620, opened 2020-01-09
Reserve "Travel Discovery" as a term that doesn't need to translate on pontoon


youwenliang:
As title.
Per our discussion we need to keep "Travel Discovery" as a feature name for all countries so don't need to translate it on pontoon.
Add this ticket for tracking.

benjamin-cheng:
@youwenliang @Conjuror 

We should follow the procedure to safely update the string in the existing code

- [ ] PM adds the new strings with new string ids (Don't override the existing strings)
- [ ] Engineer modify the code to reference the new string ids
- [ ] PM removes those strings which have been replaced and no longer used
- [ ] EPM team files l10n PR

tinahsieh:
Hey @benjamin-cheng, no prob, I'll work on it :)

Conjuror:
https://github.com/mozilla-l10n/zerda-android-l10n/pull/63

Conjuror:
18876

benjamin-cheng:
Checked with @tinahsieh, the whole requirement is `Travel Discovery` should not be translated from Protoon, however, let marketing team decide the localized translation instead. Refer to the issue #4812.

PR #4769, opened 2020-02-27
Add 'travel_search_engine_2' string resouce back to avoid build break

benjamin-cheng:
We removed the `travel_search_engine_2` from string resources, however, the string is not removed from Pontoon that required a PR to l10n repository. Also, there is a cron job everyday to import latest l10n string from Pontoon, that causes the lint violation - having string resources in specific language but not existed in "main". We'll discuss the string removal process with release manager, and currently add the string back to avoid build break.


benjamin-cheng:
Related #4620

PR #4770, opened 2020-02-27
Reserve the last "travel discovery"

tinahsieh:


app/src/main/res/values/strings.xml
```diff
@@ -991,6 +991,8 @@
     <string name="travel_dialog_2_title_new">Enjoy Using %1$s?</string>
     <!--dialog description for introducing how it works. The %1$s here is Firefox Lite.-->
     <string name="travel_dialog_2_description">You can set %1$s Travel Disovery as your default search engine for travel.</string>
+    <!--dialog description for introducing how it works. The first %1$s is Firefox Lite, and the %2$s is Travel Discovery.-->
+    <string name="travel_dialog_2_description">You can set %1$s %2$s as your default search engine for travel.</string>
```
benjamin-cheng: Should add with a new string resource id, ex. `travel_dialog_2_description_new` when the placeholders have been changed.

benjamin-cheng:
Related to #4620

PR #4785, opened 2020-03-04
Rewrite comment and mark old strings

tinahsieh:
1. Rewrite a comment based on flod's suggestion
2. Mark old strings as !NOT USE! in comments and untranslatable.

CC @chianger


benjamin-cheng:
Related to #4620

PR #4849, opened 2020-03-26
Downgrade to appcompat 1.1.0-rc1 to avoid webview crash on Android 5 devices

benjamin-cheng:
https://issuetracker.google.com/issues/141132133


benjamin-cheng:
Fix #4850

PR #4864, opened 2020-04-05
[Shopping Search] Move the settings button in shopping search to the end

### LINKED ISSUE #4827: [Shopping Search] Move the settings button in Switch UI to the end
Currently, the settings button in Switch UI is fixed on the right side of the tab bar, will need to move it to the end of the tab bar, only accessible when users scroll till the end, and change the icon. (And also remove the overlay onboarding for the settings button)

Meta bug: #4675

benjamin-cheng:
Close #4827


app/src/main/java/org/mozilla/rocket/shopping/search/ui/ShoppingSearchResultTabFragment.kt
```diff
@@ -250,6 +252,17 @@ class ShoppingSearchResultTabFragment : Fragment(), ContentTabViewContract, Back
         })
     }
 
+    private fun calculateScrollXForTab(position: Int, positionOffset: Float): Int {
```
benjamin-cheng, line 255: Reference the implementation from TabLayout:
https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/tabs/TabLayout.java#L1784

app/src/main/java/org/mozilla/rocket/shopping/search/ui/ShoppingSearchResultTabFragment.kt
```diff
@@ -251,6 +252,7 @@ class ShoppingSearchResultTabFragment : Fragment(), ContentTabViewContract, Back
                 override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) = Unit
 
                 override fun onPageSelected(position: Int) {
+                    tab_layout_scroll_view.scrollTo(calculateScrollXForTab(position, 1F), 0)
```
proevan, line 252: should we use smooth scrolling?

benjamin-cheng: Yes, I've added the scroll animation from TabLayout.

Issue #4868, opened 2020-04-07
[Smart Shopping Search] The bottom of the website is covered by the browser chrome


tinahsieh:
**Describe the bug**
The bottom of the website is always covered by browser chrome.

**Steps To Reproduce**
1. Go to Smart Shopping Search's Switch UI
2. Scroll down in any websites that have a navigation bar at the bottom (e.g. PChome/ momo購物網)
3. Found the website's navigation bar is covered.

**Expected behavior**
To see the entire website including the navigation bar.

**Actual behavior**
Half of the nav bar is covered.

**Screenshots/videos**
![Screenshot_20200407-152144_Firefox Lite Preview](https://user-images.githubusercontent.com/17559807/78658576-e5c1da80-78fc-11ea-87c8-b3bc6ae333c3.jpg)

**Reproduce rate**
100%

chianger:
Need verification.

Daisy-pliu:
verify:✅
version: Release 2.1.14(19257)
device: Samsung 8
![Screenshot_20200410-082542_Firefox Lite](https://user-images.githubusercontent.com/44990878/78954002-52abbf00-7b0d-11ea-899a-7e4ed1d20364.jpg)

PR #4880, opened 2020-04-09
fix ViewPager bottom padding in Coordinatorlayout to fix webview bott…

proevan:
…om got cutted issue


benjamin-cheng:
Thanks @proevan. It can fix the #4868 on device with display cutoffs

PR #4912, opened 2020-04-22
Popup keyboard automatically when a user adds a new tab

### LINKED ISSUE #4862: [Feature] Popup keyboard automatically when a user adds a new tab
The “+” in tab tray should go back to the home UI with the keyboard opened. We assume that when a user wants to add a new tab they would want to do a search query so having the keyboard up would be easier for them to do so.

(But if a user taps "home" button then the keyboard won't be opened automatically)

benjamin-cheng:
Close #4862

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/home/HomeFragment.kt
+++ b/app/src/main/java/org/mozilla/rocket/home/HomeFragment.kt
@@ -7,6 +7,7 @@ import android.content.Intent
 import android.graphics.Color
 import android.net.Uri
 import android.os.Bundle
+import android.os.Looper
 import android.view.GestureDetector
 import android.view.Gravity
 import android.view.LayoutInflater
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/rocket/home/HomeFragment.kt
```diff
@@ -500,6 +501,14 @@ class HomeFragment : LocaleAwareFragment(), ScreenNavigator.HomeScreen {
         homeViewModel.showContentHubClickOnboarding.observe(viewLifecycleOwner, Observer { couponName ->
             showRequestClickContentHubOnboarding(couponName)
         })
+        homeViewModel.showKeyboard.observe(viewLifecycleOwner, Observer {
+            Looper.myQueue().addIdleHandler {
+                if (!isStateSaved) {
```
proevan, line 506: why do we need to check the `isStateSaved`?

benjamin-cheng: In case user tap back leave the home fragment right after clicking new tab button in tab tray.

PR #4927, opened 2020-04-25
limit string length for firebase's user properties

proevan:
according to the documentation from firebase:
UserProperty values can be up to 36 characters long.
https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.UserProperty


app/src/main/java/org/mozilla/rocket/abtesting/LocalAbTesting.kt
```diff
@@ -31,6 +31,11 @@ object LocalAbTesting {
         activeExperiments.flatMap { it.buckets }
                 .filter { it.bucket_range_start <= userGroup && userGroup <= it.bucket_range_end }
                 .map { it.experiment_name }
+                .also { experimentName ->
+                    if (experimentName.any { it.length > 36 }) {
```
benjamin-cheng, line 35: Suggest to define the maximum number as a constant.

PR #4933, opened 2020-04-27
[News] Dailyhunt - adjust tracking and API response

### LINKED ISSUE #4715: [News] Dailyhunt - adjust tracking and API response
**Describe the bug**
Based on Dailyhunt's feedback: "For ComScore attribution, there is a small update. Now you can fetch the Url from the API dynamically as a part of items API response.
Refer documentation"

Let's see if we can use the updated API response for the tracking mechanism.

benjamin-cheng:
Close #4715


app/src/main/java/org/mozilla/rocket/content/news/data/dailyhunt/DailyHuntNewsRemoteDataSource.kt
```diff
@@ -86,6 +88,7 @@ class DailyHuntNewsRemoteDataSource(
                 val body = it.body.string()
                 val nextPageKey = PageKey.PageUrlKey(parseNextPageUrl(body))
                 val items = fromJson(body)
+                trackItemsShown(items)
```
proevan, line 91: we could update this to be called in the other thread in the future

Issue #4837, opened 2020-03-20
Make browser toolbar collapsible


youwenliang:
In normal browsing mode and private browsing mode, make toolbar collapsible when users are scrolling the content.

UX Spec: https://mozilla.invisionapp.com/share/ABWI1559QRJ#/410131596_0-0_Cover

shsu571:
Reason for this sprint:
Providing better users browsing experience.
Browsing related feature are highly correlated to retention and first day experience

PR #4952, opened 2020-05-04
support collapsible toolbar & urlbar in browser page

### LINKED ISSUE #4917: [Feature] Landscape mode with collapsible toolbar & urlbar
With the toolbar in landscape browsing mode, enable toolbar and urlbar to collapse (auto hide) when users are scrolling through contents. The behaviors will be the same in portrait mode, including showing toolbar back when visiting other urls.

This is related to #4860 

UX Spec: https://mozilla.invisionapp.com/share/ABWI1559QRJ#/screens/414477343_3-3_Scrolling_In_Landscape_Mode

proevan:
to fix #4917 and #4837


app/src/main/res/layout/fragment_browser.xml
```diff
@@ -109,14 +109,6 @@
         android:layout_height="@dimen/fixed_menu_height"
         android:layout_alignParentBottom="true"/>
 
-    <org.mozilla.rocket.nightmode.themed.ThemedView
```
benjamin-cheng, line 112: Why the divider can be removed?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -2,1518 +2,1277 @@
...
+                Observer { (isEnabled) -> setNightModeEnabled(isEnabled) })
+    }
+
+    private fun setupBottomBar() {
+        browser_bottom_bar.setOnItemClickListener(object : BottomBar.OnItemClickListener {
+            override fun onItemClick(type: Int, position: Int) {
+                when (type) {
+                    BottomBarItemAdapter.TYPE_TAB_COUNTER -> {
+                        chromeViewModel.showTabTray.call()
+                        showTabTrayToolbar(WEBVIEW, position)
```
benjamin-cheng, line 449: Prefer to keep `TelemetryWrapper.showTabTrayToolbar()` so we have better understanding that it does the telemetry thing.

PR #4958, opened 2020-05-07
Enable ABI splits

Dexterp37:
This PR enables ABI splits on the Firefox Lite project, generating one APK per target architecture in addition to a "universal APK" containing all the targets.

You can safely ignore the first commit: it was just me adding a convenience method to specify build settings in the `local.properties` since that's more convenient on Windows.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -39,7 +39,7 @@ android {
             generatedDensities = []
         }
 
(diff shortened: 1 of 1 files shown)


Dexterp37:
**Important note**: this would also require changing the version codes, like [Fenix is doing](https://searchfox.org/mozilla-mobile/source/fenix/buildSrc/src/main/java/Config.kt#93)

PR #4961, opened 2020-05-11
Landscape mode with search actions

### LINKED ISSUE #4914: [Feature] Landscape mode with search actions
With the toolbar in landscape browsing mode, enable users to use search button and miniURL bar to start a search action.

This is related to #4860 

UX Spec: https://mozilla.invisionapp.com/share/ABWI1559QRJ#/414477341_3-1_Principle_And_Priority

benjamin-cheng:
Close #4914


app/src/main/java/org/mozilla/focus/navigation/ScreenNavigator.kt
```diff
@@ -192,38 +200,34 @@ open class ScreenNavigator(activity: HostActivity?) : DefaultLifecycleObserver {
     }
 
     interface Provider {
-        val screenNavigator: ScreenNavigator
+        fun getScreenNavigator(): ScreenNavigator?
```
proevan, line 217: should it be nullable?

app/src/main/java/org/mozilla/focus/navigation/TransactionHelper.kt
```diff
@@ -181,25 +183,24 @@ internal class TransactionHelper(private val activity: HostActivity) : DefaultLi
     }
 
     private fun getEntryTag(entry: FragmentManager.BackStackEntry): String {
-        val result = entry.name!!.split(ENTRY_TAG_SEPARATOR).toTypedArray()
-        return result[0]
+        return entry.name?.let {
+            it.split(ENTRY_TAG_SEPARATOR).toTypedArray()[0]
+        } ?: ""
```
proevan, line 188: should we throw an error or return a default value when `entry.name` is null

benjamin-cheng: It will return "" when `entry.name` is null

app/src/main/java/org/mozilla/focus/navigation/ScreenNavigator.kt
benjamin-cheng: It's not necessary to be nullable. I'll change it.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -591,7 +591,7 @@ class BrowserFragment : LocaleAwareFragment(), BrowserScreen, LifecycleOwner, Ba
 
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
-        updateBottomBarHeight()
+        updateBottomBarLayout()
```
proevan, line 594: is there any way the custom view can be aware of the orientation changing event?

Issue #4964, opened 2020-05-14
[Feature] D0: Set default browser -Android >=7 1/4


shsu571:
This enhancement aims to increase the number of users to set Firefox Lite as their default browser, and make users increase their engagement and continue using Firefox Lite as their main browser. 
Need to apply >= Andriod 7

Detail of Spec
Doc: https://docs.google.com/document/d/1kZXiupZZr-D8a7xtEmgXipUtNV6R68XuRNYvOPwygqY/edit#

UX Spec: https://mozilla.invisionapp.com/share/3EX875TVXCY#/screens/417096800
https://mozilla.invisionapp.com/share/3EX875TVXCY#/screens/417096801

Full UX spec :https://mozilla.invisionapp.com/share/3EX875TVXCY
UI Spec: https://zpl.io/VYg1vRJ

Metrics
https://docs.google.com/spreadsheets/d/15Pqqx1J4HovyYdf_xncsPApucqZFvMJALTvVOGkBTyo/edit#gid=888392947

Data Review
https://docs.google.com/document/d/1A8wNPWw8SHReH7U0sBilBmfKvRFTBQjo0M6ITVBolyU/edit#

benjamin-cheng:
Assume the t-shirt size would be `L` with typical telemetry requirement. I may update the size when the data spec is ready.

yshlin:
Data spec done:
https://docs.google.com/spreadsheets/d/15Pqqx1J4HovyYdf_xncsPApucqZFvMJALTvVOGkBTyo/edit#gid=888392947

Daisy-pliu:
### [TC0098] Make default browser - given no default user and then user set Lite as default
**Test case** 
1. given no default browser
2. go to settings
3.Tap [set as default browser]
4. Tap [go to settings]
5. Back to Lite

**Test result**
3. seet dialog [set lite as default in 30 sec] with [cacnel] and [go to settings]
4. See Firefox Lite settings page
5. see [thanks for choosing lite]


### [TC0099] Make default browser - given chrome is default user set Lite as default 
**Test case** 
1. given no default browser
2. go to settings
3. Tap [set as default browser]
4. Tap [ok, got it]

**Test result**
2. See Firefox Lite settings page
3. See dialog [select Firefox Lite, always ] with [ok] [cancel]
4. See toast [thanks for choosing lite]


### [TC00100] Failed to make default browser - given no default 
**Test Case**
1. given no default browser
2. go to settings
[...]

PR #5004, opened 2020-05-26
Prework for the new set-default-browser flow

benjamin-cheng:
Related to #4964


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.kt
```diff
@@ -40,60 +38,59 @@ class SettingsFragment : PreferenceFragment(), OnSharedPreferenceChangeListener
...
-            val firebase = getFirebase()
-            if (firebase != null) {
-                testingId += """
-                    ${firebase.getFcmToken()}
-
-
-                    """.trimIndent()
-            }
-            testingId += TelemetryHolder.get().clientId
+            val testingId = "${getFirebase().getFcmToken()}\n\n${TelemetryHolder.get().clientId}"
```
proevan, line 93: getFirebase() could be null?

benjamin-cheng: `getFirebase()` is returned non-null `firebaseContract` object which could be `FirebaseImp` or `FirebaseNoOpImp`.

PR #5016, opened 2020-05-27
Integrate new search engine Giggle with debug build

hsiangyu69:
The logo of Giggle should be replaced


app/src/androidTest/java/org/mozilla/focus/activity/SwitchSearchEngineTest.java
```diff
@@ -98,8 +100,8 @@ public void switchSearchEngine_searchViaSearchEngineAccordingly() {
             // Check is url contains search keyword
             AndroidTestUtils.urlBarContainsText(SEARCH_KEYWORD);
 
-            // Check is url contains search engine name
-            AndroidTestUtils.urlBarContainsText(searchEngineName[0].toLowerCase());
+            // Check is url contains search engine host name
+            AndroidTestUtils.urlBarContainsText(getHostName(searchEngine.getSearchForm()));
```
proevan, line 104: why do we check the host with `www` ignored instead of entire `host`?

hsiangyu69: because while running the UI tests, we use the mock files, the URL is www.mock....

Issue #4988, opened 2020-05-20
Firefox Lite Q2 Healthy check - App performance Planning


shsu571:
For building new version, want to check the app speed of all app.

cnevinc:
After consulting with the Fenix Performance team(@mcomella), here are some action items from the engineering side:
1. [On demand Benchmark app performance](https://stackoverflow.com/questions/24307596/how-can-i-add-flavors-in-a-module-with-android-studio)
2. [In-the-wild monitoring using Firebase Monitoring in Nightly](https://firebase.google.com/docs/perf-mon/get-started-android) with [custom build flavor](https://stackoverflow.com/questions/24307596/how-can-i-add-flavors-in-a-module-with-android-studio)
3. Profile the possible bottlenecks 
4. Using [Nimbledroid](https://nimbledroid.com/my_apps) in CI
5. Leverage the Fenix Performance team(@mcomella)'s benchmark baseline tool.

cnevinc:
The goal of this issue:
1. Set up a performance baseline.
2. Find target features/components we can work on to improve the performance.

For 1. I previously want to [benchmark](https://developer.android.com/studio/profile/benchmark) the performance of creating a new  Home Screen. But in the end, I found it's quite fast since it's pure Java UI. To benchmark the target feature, we need to modularize it first. This will take extra effort. We should profile our code first then benchmarking later. 

For 2.
Compare with our competitor, difference in Battery/Stability with peers is < 1%
For App Startup Time/Render, we are 4.5%~6% slower than our competitor.  I believe there are still places to grow. Especially warm startup on low-end devices( Samsung Galaxy A20, J6, M20, M30). 

My conclusion is to add Firebase Monitoring in Nightly. And find out the target we want to profile/fine-tune. There are many [tools](https://developer.android.com/topic/performance/tracing) out there for us to dig deeper. Here are my assumptions:
In the feature dimension, we can look at:
1. First Launch Experience (Cold Start, including open from external apps)
2. Warm Start (user switch back to our app)
3. Rendering (using Firebase monitoring to set the baseline for each screen)

In the user dimension, we can look at
1. Devices with only 1 to 2 cores.
2. Devices with low memory (2 to 4 GB)
3. With low network latency (which may affect fetching Remote Config, Content Services).

perceived performance
1. Context Menu
2. Starting a download 
3. Saving a bookmark
4. Open Tab Tray
5. Launch BrowserFragment.

chianger:
[...]

PR #5024, opened 2020-06-01
Enable Firebase Monitoring in Nightly

cnevinc:
related to #4988
I'll like to merge Firebase performance monitoring in Nightly only and collect some key performance metrics (e.g. cold start)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,6 +8,7 @@ apply from: '../buildSrc/pmd.gradle'
 apply from: '../buildSrc/checkstyle.gradle'
 apply from: '../buildSrc/findbugs.gradle'
 apply from: 'buildscripts/l10n.gradle'
+apply plugin: 'com.google.firebase.firebase-perf'
 
 android {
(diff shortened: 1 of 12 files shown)


cnevinc:
The last commit is a little tricky.
com.firebase.iid.zzbb that perf-monitoring brought in's version is 20.2.0. and it's final
and since com.firebse.mssage.zzf needs to implement that class, r8 will fail when running gradle task `:app:dexBuilderFocusWebkitDebug` 
```
D8: Type com.google.firebase.iid.zzbb is referenced as an interface from `com.google.firebase.messaging.zzf`.
```
There are two ways to fix this. Using an older version of perf-monitoring. or use a newer version of FCM. I choose the latter since we'll need to upgrade it anyway in the future. It's better we deal with it now.

I'll have another PR to upgrade other FCM libraries as well.  Cause some libraries are calling deprecated methods (e.g. Remote Config), and Crashlytics (migrated from Fabric to FirebaseCrashlytics )

PR #5052, opened 2020-06-02
Set default browser (Android >=7)

### LINKED ISSUE #4964: [Feature] D0: Set default browser -Android >=7 1/4

benjamin-cheng:
Close #4964


app/src/main/java/org/mozilla/rocket/settings/defaultbrowser/ui/DefaultBrowserPreferenceViewModel.kt
```diff
@@ -39,27 +55,62 @@ class DefaultBrowserPreferenceViewModel(private val defaultBrowserRepository: De
...
-                triggerWebOpen.call()
+                openUrlTutorialDialog.call()
+                TelemetryWrapper.showSetDefaultBySettingsMessage()
             }
         }
     }
 
     fun onResume() {
         refreshSettings()
+        if (tryToSetDefaultBrowser) {
```
cnevinc, line 74: Nit: I'll suggest return early to make it less nested :)

app/src/androidTest/java/org/mozilla/focus/activity/DefaultBrowserTest.java
```diff
@@ -126,6 +126,11 @@ public void changeDefaultBrowser_whenChromeIsDefault() {
         // Click on "Default Browser" setting, this will brings up the Android system setting
         clickDefaultBrowserSetting(prefName);
 
+        // Click on the positive button of the tutorial dialog to continue
```
cnevinc, line 129: Normally if the tutorial is not critical to the user or will change over time, I'll suggest we just skip it in the UI test
For example, change the preference before we launch the UI test.
But since we didn't use preference, instead we use PackageManager to check, let's keep it as is.

PR #5061, opened 2020-06-02
Support deeplink to open settings page

hsiangyu69:


app/src/main/java/org/mozilla/rocket/deeplink/DeepLinkConstants.kt
```diff
@@ -18,4 +19,4 @@ object DeepLinkConstants {
 
     const val COMMAND_PARAM_KEY = "command"
     const val COMMAND_SET_DEFAULT_BROWSER = "setDefaultBrowser"
-}
```
cnevinc, line 21: Nit: Extra new line

hsiangyu69: When trying to push I will get this error app/src/main/java/org/mozilla/rocket/deeplink/DeepLinkConstants.kt:1:1: File must end with a newline (\n)

PR #5066, opened 2020-06-04
Monitor news items shown from UI layer

### LINKED ISSUE #4985: [Dailyhunt] Track request is sent for all items even before they came into the user's view
Per Dailyhunt's request:

1. Immediately on App launch the client sends 6-8 requests to fetch items for the first and second tab enabled in the browser and sometimes items worth of 3 pages fetched in one go with pageSize=30. Please note that this is not the intended behaviour.
The APIs are fast, so preloading of content in tabs is not needed. In "For You" channel all three pre-loaded pages are added to user's history whereas the items were never viewed by the user.
As a result, the user might start seeing old content in the next refresh or launch.

2. Along with content pre-loading, the track request is sent for all those items, even before they came into the user's view.
This will also result in tracking the items not seen by the user. Also skewing the Page view numbers.
The track request should only be sent for only the items which came into the user's view. In current implementation, the user views only 6-7 items in the list, but is tracked against 50-60 items. That's 6 page views being tracked where it should've been 1.

3. The refresh button refreshes all tabs irrespective of whether the user has viewed the content in them or not. This was highlighted in the integration phase as well.
This also results in content never seen by user but yet tracked.

benjamin-cheng:
Close #4985

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/content/common/ui/VerticalTelemetryViewModel.kt
+++ b/app/src/main/java/org/mozilla/rocket/content/common/ui/VerticalTelemetryViewModel.kt
@@ -1,6 +1,8 @@
 package org.mozilla.rocket.content.common.ui
 
 import android.os.Looper
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mozilla/rocket/content/common/ui/VerticalTelemetryViewModel.kt
```diff
@@ -46,11 +60,18 @@ class VerticalTelemetryViewModel : ViewModel() {
     fun updateImpression(category: String, subCategoryId: String, maxIndex: Int) {
         val index = impressionMap[category]?.get(subCategoryId) ?: 0
         if (index < maxIndex) {
+            val isFirstImpression = (impressionMap[category] == null)
             if (impressionMap[category] == null) {
```
proevan, line 64: `if (isFirstImpression)`

PR #5093, opened 2020-06-09
Menu items and add top sites

tinahsieh:
- menu items
- add top sites toast and snackbar
- Refine the title of "set default browser"


tinahsieh:
@Daisy-pliu I updated the copy string! Ready for Pontoon :)

PR #5167, opened 2020-06-30
Add topsite from recommeded site panel - success UI flow

### LINKED ISSUE #5048: [Home customization] Top site 3/7 (Add Topsite)
**User story**
As a user, I want to be able to customize the top site area, such as pinning webpage to topsite area or entering url to add to topsite area or reordering the topsites or removing the pinned sites, so that I can display websites in my own terms. (And don’t drag browsing history to this area)

**Additional context**
When users add any topsite from the Topsite selection panel, add new topsite will pop up with a snackbar prompting users to add more topsite. Will also need to consider when all topsite area is full, or all topsites are removed.

- If users remove browsing data:
1. Pin topsite won't be removed.
2. Topsites that came with presets won't be removed.
3. Topsites that came with presets but were removed by users will stay removed.
4. Topsites that pop-up by frecency will be removed.

Logic: https://mozilla.invisionapp.com/share/7YXG8U1NEJZ#/419234362_4-1_Top_Site_Placing_Logic

Overall UX Spec: https://mozilla.invisionapp.com/share/7YXG8U1NEJZ#/419234356_3-1_Add_New_Top_Sites

Visual Spec https://zpl.io/a7AGrBR

benjamin-cheng:
Close #5048

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
@@ -72,6 +72,7 @@ import org.mozilla.rocket.content.getViewModel
 import org.mozilla.rocket.download.DownloadIndicatorViewModel
 import org.mozilla.rocket.extension.nonNullObserve
 import org.mozilla.rocket.home.HomeFragment
+import org.mozilla.rocket.home.topsites.ui.AddNewTopSitesActivity
 import org.mozilla.rocket.landing.DialogQueue
 import org.mozilla.rocket.landing.NavigationModel
 import org.mozilla.rocket.landing.OrientationState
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mozilla/rocket/chrome/ChromeViewModel.kt
```diff
@@ -86,6 +87,7 @@ class ChromeViewModel(
     val openPreference = SingleLiveEvent<Unit>()
     val showFindInPage = SingleLiveEvent<Unit>()
     val showAdjustBrightness = SingleLiveEvent<Unit>()
+    val addNewTopSiteSuccess = SingleLiveEvent<Int>()
```
proevan, line 90: we can move it to HomeViewModel

benjamin-cheng: fixed

PR #5203, opened 2020-07-06
adjust menu item height and menu height

### LINKED ISSUE #5067: [Home Customization] Menu 3/3 (Shopping Search/Private Mode)
**User story**
- As a user, I want to be able to make further customization such as changing the individual items (Verticals, Searchbar, topsite area, shopping search button...etc.) on the Home screen, so that I can access the features I want or hide the features I don’t need in order to improve my browsing experience.

**Summary**
If users have Private mode as their feature button, replace private mode in menu to shopping search, vise versa.

**Additional context**
Group

proevan:
close #5067

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -22,6 +22,8 @@
     <dimen name="menu_corner_radius">16dp</dimen>
     <dimen name="menu_horizontal_margin">4dp</dimen>
     <dimen name="menu_bottom_margin">12dp</dimen>
+    <dimen name="menu_item_size">48dp</dimen>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/dimens.xml
```diff
@@ -22,6 +22,8 @@
     <dimen name="menu_corner_radius">16dp</dimen>
     <dimen name="menu_horizontal_margin">4dp</dimen>
     <dimen name="menu_bottom_margin">12dp</dimen>
+    <dimen name="menu_item_height">48dp</dimen>
```
benjamin-cheng, line 25: Suggest to use a neutral term since it represents both item width and height ex. `menu_item_size`

PR #5206, opened 2020-07-06
Cancel the swipe down gesture for dismissing menu. closes #5113

### LINKED ISSUE #5113: [Menu] Cancel the swipe down gesture for dismissing menu
**User story**
Since the swipe down gesture can be activated very easily, we'll test it out with more users to see if we have to cancel the gesture.

Version A: **Swipe down** or **click the overlay** to dismiss the menu
Version B: **Click the overlay** to dismiss the menu

cnevinc:


app/src/main/java/org/mozilla/rocket/menu/BrowserMenuDialog.kt
```diff
@@ -71,6 +71,8 @@ class BrowserMenuDialog : BottomSheetDialog {
 
         initLayout()
         observeChromeAction()
+        setCancelable(false)
+        setCanceledOnTouchOutside(true)
```
proevan, line 75: also need to add in ListPanelDialog

PR #5208, opened 2020-07-06
[Strings] Replace the "." in the string ID

tinahsieh:


tinahsieh:
Updated the strings by adding the previous strings with the ID "2.5".

PR #5211, opened 2020-07-07
Separate night mode and dark theme

### LINKED ISSUE #5157: [Home customization] Onboarding  (Android dark theme)
Redefine the current night mode to be only adjusting the screen brightness:
Dark theme: UI -> Black (disconnect the brightness adjustment)
Night mode: brightness adjustment (disconnect the UI theme setting)

hsiangyu69:
Close #5157


app/src/main/java/org/mozilla/rocket/home/HomeViewModel.kt
```diff
@@ -93,6 +94,7 @@ class HomeViewModel(
     val isFxAccount: LiveData<Boolean> = getIsFxAccountUseCase()
     val shouldShowNewMenuItemHint: LiveData<Boolean> = shouldShowNewMenuItemHintUseCase()
     val isContentHubEnabled: LiveData<Boolean> = shouldShowContentHubUseCase()
+    val homeBackgroundColorTheme = MutableLiveData<ThemeManager.ThemeSet>()
```
proevan, line 97: we should probably initialize the value at the beginning. But we might able to do it in the future.

PR #5237, opened 2020-07-14
[Strings] New in-app survey

tinahsieh:
1. Update the existing survey link
2. Add a new in-app survey with dialog copy strings and survey link


shsu571:
same as https://github.com/mozilla-tw/FirefoxLite/issues/5235

PR #5240, opened 2020-07-15
remove content hub onboarding

proevan:


app/src/main/java/org/mozilla/focus/utils/DialogUtils.kt
```diff
@@ -313,40 +313,6 @@ object DialogUtils {
...
 
-    fun showContentServiceOnboardingSpotlight(
-        activity: FragmentActivity,
-        targetView: View,
-        dismissListener: DialogInterface.OnDismissListener,
-        ok: View.OnClickListener?
-    ): Dialog = SpotlightDialog.Builder(activity, targetView)
-            .spotlightConfigs(
-                RectangleSpotlightConfigs(
-                    width = activity.resources.getDimensionPixelSize(R.dimen.content_service_focus_view_width),
```
benjamin-cheng, line 324: Could we also remove the `content_service_focus_view_width` and `content_service_focus_view_radius` definitions from `dimens.xml`?

app/src/main/res/layout/onboarding_spotlight_content_services_logo_man.xml
```diff
@@ -1,5 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<ImageView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
-    android:src="@drawable/logo_man_copy_3"/>
```
benjamin-cheng, line 5: Could we remove the `logo_man_copy_3` asset?

app/src/main/java/org/mozilla/focus/utils/DialogUtils.kt
proevan: those two assets are still used in DialogUtils.showContentServiceRequestClickSpotlight

app/src/main/res/layout/onboarding_spotlight_content_services_logo_man.xml
proevan: fixed in the last commit, thanks for mention

PR #5244, opened 2020-07-15
keep the second page of the top sites area

### LINKED ISSUE #5236: Keep the second page of the top site area
Since right now users can long press the empty space to add top site, we encounter an issue when users have 8 topsites in the first page, there's no empty space for them to long press.

We propose to keep the second page all the time even when users have less than 8 topsites.
This way when there are already 8 topsites, they can swipe to the second page (which is emptied) and long press to add more topsites.

hsiangyu69:
Close #5236

### DIFF
--- a/app/src/main/java/org/mozilla/rocket/home/HomeViewModel.kt
+++ b/app/src/main/java/org/mozilla/rocket/home/HomeViewModel.kt
@@ -194,7 +194,7 @@ class HomeViewModel(
     private fun updateTopSitesData() = viewModelScope.launch {
         val topSiteList = getTopSitesUseCase()
         sitePages.value = if (topSiteList.isNotEmpty()) {
-            topSiteList.toSitePages().addDummyTopSites().also { sitePages ->
+            topSiteList.addDummyTopSites().toSitePages().also { sitePages ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/rocket/home/HomeViewModel.kt
```diff
@@ -212,28 +212,21 @@ class HomeViewModel(
...
-                for (i in 0 until TOP_SITES_PER_PAGE - sitePage.sites.size) {
-                    sites.add(Site.DummySite)
-                }
-                sitePagesWithDummySite.add(SitePage(sites))
-            } else {
-                sitePagesWithDummySite.add(sitePage)
-            }
+    private fun List<Site>.addDummyTopSites(): List<Site> {
+        val siteListWithDummySite = this.toMutableList()
+        val dummySiteSize = TOP_SITES_PER_PAGE * 2 - this.size
```
proevan, line 225: we can replace `2` with `TOP_SITES_MAX_PAGE_SIZE`

hsiangyu69: Sure ! thank you

PR #5253, opened 2020-07-15
Set-default-browser onboarding

### LINKED ISSUE #5037: [Home customization] Onboarding 5/5 (Completion)
**User story**
- As a new user, I want to be able to feel a sense of completion after customizing my home experience, and be able to have an option to set Firefox Lite as my default browser so that I can used my customized Firefox Lite in the future.

**Additional context**
UX Spec: https://mozilla.invisionapp.com/share/7YXG8U1NEJZ#/419234351_1-1_New_User_Onboarding_Flow_-2-2-
Visual spec: https://zpl.io/2yA0nzG

benjamin-cheng:
Close #5037


app/src/main/java/org/mozilla/focus/utils/DialogUtils.kt
```diff
@@ -468,7 +468,7 @@ object DialogUtils {
         ThemeSettingDialogBuilder(activity, homeViewModel).show()
     }
 
-    fun showSetAsDefaultBrowserDialog(activity: FragmentActivity) {
+    fun showSetAsDefaultBrowserDialog(activity: FragmentActivity, homeViewModel: HomeViewModel) {
```
proevan, line 471: how about pass a callback instead of viewmodel?

app/src/main/java/org/mozilla/rocket/theme/ThemeSettingDialogBuilder.kt
```diff
@@ -29,6 +29,9 @@ class ThemeSettingDialogBuilder(
 
         return AlertDialog.Builder(activity, R.style.ThemeSettingDialog)
             .setView(customContentView)
+            .setOnDismissListener {
+                homeViewModel.onExitThemeSetting()
```
proevan, line 33: I'll also consider to pass callback into this class to handle the events

app/src/main/java/org/mozilla/focus/utils/DialogUtils.kt
benjamin-cheng: Fixed

app/src/main/java/org/mozilla/rocket/theme/ThemeSettingDialogBuilder.kt
benjamin-cheng: Per discussed, no need to change this one

PR #5310, opened 2020-07-31
Keep file extension as is.

cnevinc:


cnevinc:
> 👍
> Do you have the download url that would return ".bin" from `guessFileName`?
Here:
https://msrp-dev01.appspot.com/api/v1/admin/push/ui?token=3
image

or just click:
https://msrp-dev01.appspot.com/img/download

Issue #5228, opened 2020-07-13
Download Management stage 0 


shsu571:

benjamin-cheng:
1. [ ] Remove callbacks from async content queries
2. [ ] Move EnqueueDownloadTask to DownloadDataSource
3. [ ] Move GetDownloadFileHeaderTask to DownloadDataSource
4. [ ] Move download broadcast receiver to repository/data source 
5. [ ] Move showOpenDownloadSnackBar from DownloadInfoManager to UI layer
6. [ ] Separate download implementation to DownloadDataSource
7. [ ] Combine RelocateService in downloading module
8. [ ] Combine DownloadCompleteReceiver in downloading module
9. [ ] Integrate new DownloadManager
10. [ ] Implement dedicate foreground process for downloading service

PR #5316, opened 2020-08-04
Refactor download management - 2

benjamin-cheng:
Related to #5228


app/src/main/java/org/mozilla/rocket/download/data/DownloadInfoManager.kt
```diff
@@ -90,8 +90,19 @@ class DownloadInfoManager {
         )
     }
 
-    fun updateByRowId(downloadInfo: DownloadInfo, listener: AsyncUpdateListener?) {
-        mQueryHandler.startUpdate(TOKEN, listener, DownloadContract.Download.CONTENT_URI, getContentValuesFromDownloadInfo(downloadInfo), DownloadContract.Download._ID + " = ?", arrayOf(downloadInfo.rowId.toString()))
+    suspend fun updateByRowId(downloadInfo: DownloadInfo) = suspendCoroutine<Int> { continuation ->
```
proevan, line 87: should we do the query in IO thread?

benjamin-cheng: SQL query already runs in the background thread by AsyncQueryHandler.

PR #5349, opened 2021-01-27
Upgrade Compile Target to 29

st3fan:
This patch sets `compile_target` to `29` and includes a few small fixes for compilation issues. We also set `kotlinOptions.allWarningsAsErrors = false` to ignore deprecation warnings. At this point we will not fix those unless absolutely required.


jonalmeida (approved):
LGTM

Issue #5348, opened 2021-01-27
After the API 29 upgrade, screenshots do not work anymore on Android 10 or newer


st3fan:
After the API 29 upgrade, screenshots do not work anymore on Android 10 or newer. I'm looking at `adb logcat` but there are no exceptions and AFAICT no messages. But I could be wrong about the latter. We do get a toast saying "Screenshot failed" so maybe that is triggered by application code in a failure path.

I tried the Screenshot functionality on 2.5.1 (Which is linked against API level 28) and there it works on the same Android 11 devices that I have for testing.

st3fan:
Looks like `getPageBitmap` catches possible exceptions but does not report on those

https://github.com/mozilla-mobile/FirefoxLite/blob/f47ddacb1df7fc20a741bfbbe5356d26ef967fd9/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt#L1054

Not sure if that is where this is failing though.

st3fan:
The root cause of this turned out to be storage changes in Android 10. A fix is in #5350

PR #5350, opened 2021-01-28
Request Legacy External Storage

st3fan:
This patch adds `android:requestLegacyExternalStorage="true"` to the manifest to temporarily opt-out of scoped storage. This is something that needs to be properly fixed before the application can move to API level 30 (Android 11).

This is a fix for #5348 (After the API 29 upgrade, screenshots do not work anymore on Android 10 or newer) and requires #5349 to land first since this app settings requires build tools 29.

Thank you @grigoryk for debugging this and finding the root cause.

> Hm this patch has a bit of a weird commit history since I based it on another PR branch.


st3fan:
Closed in favor of #5351

